PR #10, opened 2015-07-29
Road to 1.0: Isolate symfony related code

dupuchba:
As mentioned https://github.com/dunglas/api-platform/issues/9 I am clearing ap-platform from custom examples.


dunglas:
I was unclear I think. The goal is to make DunglasApiBundle usable standalone, not to remove examples nor Symfony from the standard edition.

dupuchba:
Oh ok, so how about `php-schema`dev dependencies ? and nelmio ?

dunglas:
I think they are perfectly ok. The default edition must be a way to easily create API for a developer. The idea here is too make libraries usable standalone (without the standard edition in custom projects for advanced needs like DDD or integration in legacy projects).

dupuchba:
ok, so nothing to do with `api-platform` ^_^. I was confused by the fact that you wrote Road to 1.0 in `api-platform`

dunglas:
Sorry. I'll edit the issue.

PR #14, opened 2015-09-09
Fix behatch version

### LINKED ISSUE #13: Behat version and Given issue
I've got some error with basic behat scenario 
```
  @createSchema
  Scenario: Create a Person
    Given Abc
```
where Behat send a Exception
```
[Behat\Mink\Exception\DriverException]                
  Unable to access the response before visiting a page
```

It's a Behatch version problem. `dev-master` works fine where `~2.1` breaks. The Behatch version in DunglasApiBundle is already set to `dev-master`. I don't see this as a good practice but it could be a temporary solution for api-platform too.  

There's also reference to this issue [here](https://github.com/Behatch/contexts/issues/126)

theofidry:
A bug make the behat scenarios throw an error in the version 2.1.*
of behatch/contexts. Due to the removing of the later tags 2.2 and
2.3 as referenced in the issue behatch/contexts#126, the dependency
is set to dev-master for the moment.

Closes #13

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "behat/mink": "~1.5",
         "behat/mink-extension": "~2.0",
         "behat/mink-browserkit-driver": "~1.1",
-        "behatch/contexts": "~2.1",
+        "behatch/contexts": "dev-master",
         "hautelook/alice-bundle": "~0.2"
     },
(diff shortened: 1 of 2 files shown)


dunglas:
Thank you @theofidry

PR #18, opened 2015-10-05
Delete entity AppBundle\Entity\Thing

toofff:
The class "Thing" must not exist. There is no reference in the "app/config/schema.yml" file.


dunglas:
Thank you @toofff

PR #22, opened 2015-10-15
fix the removing of authors and illustrators in the Book entity

### LINKED ISSUE #21: Error removing Persons from the Book entity
If I try to remove Persons from a Book I get the following error:

```
PUT http://localhost:9000/books/1
Content-Type: application/json
{
    "name": "Book 1", 
    "illustrator": [],
    "author": []
}
 -- response --
400 Bad Request
...
 "array_search() expects parameter 2 to be array, object given",
    "api-platform/src/AppBundle/Entity/Book.php", 106
    and 
    "api-platform/src/AppBundle/Entity/CreativeWork.php", 95,
...
```

This occurs because a Doctrine\ORM\PersistentCollection is passed as a second argument.

I fixed it and made a pull request #22

lubomir-haralampiev:
fixes #21 - Error removing Persons from the Book entity

### DIFF
--- a/app/config/schema.yml
+++ b/app/config/schema.yml
@@ -3,7 +3,7 @@
 # bin/schema generate-types src/ app/config/schema.yml
 
 rdfa:
-  - https://raw.githubusercontent.com/schemaorg/schemaorg/sdo-gozer/data/schema.rdfa
+  - http://schema.org/docs/schema_org_rdfa.html
 annotationGenerators:
     - ApiPlatform\SchemaGenerator\AnnotationGenerator\PhpDocAnnotationGenerator
     - ApiPlatform\SchemaGenerator\AnnotationGenerator\DoctrineOrmAnnotationGenerator
--- a/src/AppBundle/Entity/Book.php
+++ b/src/AppBundle/Entity/Book.php
@@ -103,9 +103,8 @@ class Book extends CreativeWork
      */
     public function removeIllustrator(Person $illustrator)
     {
-        $key = array_search($illustrator, $this->illustrator, true);
-        if (false !== $key) {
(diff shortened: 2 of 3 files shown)


dunglas:
These file are automatically generated by https://github.com/api-platform/schema-generator. Can you open a PR to fix the generator first then regenerate files?

lubomir-haralampiev:
OK, I'll take a look at it

lubomir-haralampiev:
I fixed the generator https://github.com/api-platform/schema-generator/pull/32.

Though the url in the example schema https://raw.githubusercontent.com/schemaorg/schemaorg/sdo-gozer/data/schema.rdfa includes a wrong content-type 'text/plain' in its response and the rdfa can not be parsed.
After serving the same data from my local machine using the correct content-type 'application/xhtml-xml' (according to http://www.w3.org/TR/xhtml-rdfa/) I was able to generate the files successfully.
Unfortunately I can't provide a public url with the correct content-type.

How to deal with it?
In my opinion this should be another issue.

dunglas:
Yes it should. What about using this file: http://schema.org/docs/schema_org_rdfa.html?

lubomir-haralampiev:
It works.
However genre and publisher have each two ranges defined:
genre - Text and URL
publisher - Organization and Person 

This causes the TypesGenerator::generateField to log an error:
```
[error] The property "genre" (type "CreativeWork") has several types. Using the first one.
[error] The property "publisher" (type "CreativeWork") has several types. Using the first one.
```

Though all types are generated properly.

sroze:
I'm closing this PR as it's now out of date, and we should simply regenerate all the entities.

PR #25, opened 2015-11-16
Use 2.3 for behatch/contexts

blazarecki:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "behat/mink": "~1.5",
         "behat/mink-extension": "~2.0",
         "behat/mink-browserkit-driver": "~1.1",
-        "behatch/contexts": "dev-master",
+        "behatch/contexts": "^2.3",
         "hautelook/alice-bundle": "~0.2"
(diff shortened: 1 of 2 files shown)


sroze:
I would prefer to remove this dependency, and remove it from the `behat.yml` file.

dunglas:
@sroze Why? AFAIK there is no better alternative out yet.

theofidry:
Even if we do consider moving away from it, in the meantime it won't hurt to used a fixed version instead of `dev-master`.

sroze:
Yup, that won't hurt. 

@blazarecki could you rebase on `master` ?

blazarecki:
@sroze Done.

dunglas:
Thanks @blazarecki

PR #27, opened 2015-11-20
BeforeScenario and AfterScenario not called

jdeniau:
Apparently the other annotations blocks the createDatabase and dropDatabase functions to be called

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -35,7 +35,7 @@ class FeatureContext implements Context, SnippetAcceptingContext
     }
 
     /**
-     * @BeforeScenario @createSchema
+     * @BeforeScenario
      */
     public function createDatabase()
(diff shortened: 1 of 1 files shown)


jderusse:
Database shouldn't be re-created on each scenarii.

jdeniau:
Well in that case, those method should disappear ^^ (or be changed to BeforeFeature or something like that)

dunglas:
I don't understand what is the intent of this PR. The database is not created or deleted at each scenario, only on purpose with annotations.

jdeniau:
Well I think I do not understand how do your create / drop your database ?

dunglas:
Here is an example: https://github.com/dunglas/DunglasApiBundle/blob/1.0/features/crud.feature

jdeniau:
Ok, thank your and sorry for this PR ;)

PR #47, opened 2016-03-05
Update the Docker configuration

sroze:
Update the Docker configuration to remove the host-specific configuration and use [immutable images](http://sroze.io/2015/12/23/immutable-docker-images/).


Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# Install composer
+RUN curl -sS https://getcomposer.org/installer | php \
+    && mv composer.phar /usr/bin/composer
+
+# Add the application
+ADD . /app
+WORKDIR /app
+
+# Remove cache and logs if some and fixes permissions
+RUN ((rm -rf app/cache/* && rm -rf app/logs/*) || true) \
```
jderusse: What'd about using .dockerignore

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    && chown www-data . app/cache app/logs
+
+# Install dependencies
+RUN composer install -o
```
jderusse: Should be run as www-data to avoid permissions issues with files generated by scripts.

Dockerfile
sroze: `.dockerignore` is not recursive :(

jderusse:
Can run a cache warmup too

dunglas:
:+1:

Dockerfile
jderusse: Even with `**` pattern ?
I never used it, but I'm really surprised we can no ignore complete directory. :(

sroze: This was [merged](https://github.com/docker/docker/pull/17090) for Docker 1.10 only, while most of the users still don't use it.

Dockerfile
sroze: When doing that, you basically have to update your `/etc/passwd` to be able to `su - www-data`, Then, with the `.:/app` volume in the `docker-compose.yml` file, your existing files in `vendor/` do not belong to `www-data` so you need to chown everything.

Don't get me wrong I'm not against. Still, do you think that's worth doing it? I like the idea of warming up the cache though.

jderusse: May be, just move the chown www-data after the composer install?

sroze:
@jderusse I've added the cache warmup, good idea!

PR #49, opened 2016-03-10
[docker] install git + add missing extensions: zip, pdo_mysql.

aitboudad:


Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM php:7.0-apache
 
+RUN apt-get update && apt-get install -y git && apt-get install -y libpng-dev && apt-get install -y zlib1g-dev
+
 # PHP extensions
 RUN docker-php-ext-install mbstring
+RUN docker-php-ext-install mysqli pdo pdo_mysql
```
teohhanhui: This is from my Dockerfile:

```Dockerfile
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        libicu-dev \
        zlib1g-dev \
    && rm -rf /var/lib/apt/lists/* \
    && docker-php-ext-install \
        intl \
        mbstring \
        pdo_mysql \
        zip
```

We definitely don't need `mysqli`, and `pdo` is already included.

aitboudad: fixed

dunglas:
:+1: ping @sroze

Dockerfile
teohhanhui: Let's do them in a single `RUN` command. We don't need so many layers...

Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM php:7.0-apache
 
+RUN apt-get update && apt-get install -y git && apt-get install -y libpng-dev && apt-get install -y zlib1g-dev
+
 # PHP extensions
 RUN docker-php-ext-install mbstring
+RUN docker-php-ext-install pdo_mysql
```
teohhanhui: Let's do them in a single RUN command. We don't need so many layers...

Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM php:7.0-apache
 
+RUN apt-get update && apt-get install -y git && apt-get install -y libpng-dev && apt-get install -y zlib1g-dev
```
teohhanhui: What do we need `libpng-dev` for?

Dockerfile
jderusse: Why mysql and not sqlite nor pgsql

Dockerfile
jderusse: Exclude `recommands` from install packages

jderusse: Remove /etc/apt/ files after install

teohhanhui:
```Dockerfile
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        libicu-dev \
        zlib1g-dev \
    && rm -rf /var/lib/apt/lists/* \
    && docker-php-ext-install \
        intl \
        mbstring \
        zip
```

Dockerfile
teohhanhui: Good point. Let's leave that out.

p/s: I hope to contribute my Docker Compose setup.

dunglas: Currently the default setting of both API Platform and Symfony SE is MySQL. I personally prefer Postgres but MySQL is more used in the PHP world.

jderusse:
Yep, exactly. But remove `*-dev` packages after install because they are only needed for compilation. Replace it by the related lib when needed.

Dockerfile
teohhanhui: It's not very useful if mysql is not included in this Docker setup.

aitboudad:
should I include `intl` too ?

teohhanhui:
http://symfony.com/doc/current/reference/requirements.html#optional

dzubchik:
@aitboudad , what about opcache? It gives 2 times less memory consumption and better response time

teohhanhui:
@dzubchik Good idea. And we can install `apcu` too...

```Dockerfile
RUN echo "zend_extension=opcache.so" >> /usr/local/etc/php/conf.d/05-opcache.ini

RUN pecl install \
        apcu \
        apcu_bc-beta
RUN echo "extension=apcu.so" >> /usr/local/etc/php/conf.d/20-apcu.ini \
    && echo "extension=apc.so" >> /usr/local/etc/php/conf.d/21-apc.ini
```

(not sure if `apcu_bc` is necessary...)

@aitboudad @jderusse

jderusse:
Use `docker-php-ext-enable` instead of writing in `.ini` files.

`apcu-bc` is necessary when your application use `apc_*`. I didn't test recently but last versions of Symfony should be compliant.

teohhanhui:
@jderusse If I remember correctly the load order needs to be `apcu.so` followed by `apc.so`, and we have no control over that when using `docker-php-ext-enable`.

jderusse:
You can name `.ini` file with commonly used `zz-` prefix.
https://github.com/docker-library/php/blob/master/7.0/docker-php-ext-enable#L10

teohhanhui:
Didn't know that. Cool!

aitboudad:
I don't have much time to complete this pr, @teohhanhui could you take care of it ? thank's

PR #54, opened 2016-03-20
Add Fixtures

### LINKED ISSUE #53: Adding fixtures
Should we add some fixtures out of the box to allow the user to see how the application behave (and not an empty database)?

cc @dunglas

theofidry:
Closes #53 

Add 10 fixtures for each entity; to load them: `php bin/console h:f:d:l` assuming you are using Doctrine ORM.


app/AppKernel.php
```diff
@@ -26,6 +26,8 @@ public function registerBundles()
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
             $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
             $bundles[] = new Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle();
+            $bundles[] = new Hautelook\AliceBundle\HautelookAliceBundle();
+            $bundles[] = new Fidry\PsyshBundle\PsyshBundle();
```
dunglas: Is this one really stable and necessary?

composer.json
```diff
@@ -26,29 +26,32 @@
     "require": {
         "php": ">=7.0",
         "symfony/symfony": "3.1.*@dev",
+        "api-platform/core": "^2.0@dev",
         "doctrine/orm": "^2.5",
         "doctrine/doctrine-bundle": "dev-property_info",
         "doctrine/doctrine-cache-bundle": "^1.2",
-        "symfony/swiftmailer-bundle": "^2.3",
```
dunglas: Why those changes? I prefer to keep them synchronized with the Symfony Standard Edition. It makes it easy to merge changes.

app/AppKernel.php
theofidry: Psysh? Yeah it is, but I putted here as debug and forgot to remove it. It could be included by default though, it's just a debug tool. So your call

src/AppBundle/DataFixtures/Provider/GenderProvider.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace AppBundle\DataFixtures\Provider;
+
+use Faker\Provider\Base as BaseProvider;
+
+/**
+ * @author Théo FIDRY <theo.fidry@gmail.com>
```
dunglas: IMO we should not add authors in the edition to make it clear that users must appropriate and change such files. Btw, adding a provider looks too advanced for me.

composer.json
theofidry: Lexicographical order (sorry it's me being a maniac); will revert

src/AppBundle/Entity/Book.php
```diff
@@ -32,7 +32,7 @@ class Book
      * @ORM\ManyToMany(targetEntity="AppBundle\Entity\Person")
      * @Property(iri="http://schema.org/author")
      */
-    private $author;
```
dunglas: This file should be kept in sync with what the `schema.yml` file produces.

dunglas:
:+1: for the idea but I've left some comments about the implementation.

src/AppBundle/DataFixtures/Provider/GenderProvider.php
theofidry: Got it; should I use a static 'male' string for example instead?

src/AppBundle/Entity/Book.php
theofidry: ok, will do that in another PR then

dunglas:
Do you have some time to finish this one?

src/AppBundle/Entity/Book.php
```diff
@@ -84,7 +84,7 @@ class Book
     /**
      * @var Organization The publisher of the creative work.
      *
-     * @ORM\OneToOne(targetEntity="AppBundle\Entity\Organization")
+     * @ORM\ManyToOne(targetEntity="AppBundle\Entity\Organization")
```
theofidry: @dunglas not sure how you update that in you `schema.yml`

theofidry:
@dunglas updated

dunglas:
👍

src/AppBundle/Entity/Book.php
dunglas: The "cardinality" setting must be changed.

Issue #50, opened 2016-03-11
Provide a Docker Compose setup


teohhanhui:
Looking at #49, I think we should provide a complete Docker Compose setup.

Do you think it should be in a separate repo? I'll see if I can contribute mine as a start.

dunglas:
Isn't more developer friendly to have a working docker setup just after doing `composer create-project api-platform/api-platform`?

theofidry:
>Isn't it more developer friendly to have a working docker setup just after doing composer create-project api-platform/api-platform

I also think so :)

teohhanhui:
I wonder if there could be reuse of `Dockerfile`s across the different distributions (Laravel edition etc. as planned)?

We could also look into automated builds for the Docker images (if it makes sense in our case).

sroze:
We definitely need that Docker Compose configuration, true!

But I'm not sure it really makes sense for us to have an official _built_ Docker image.

teohhanhui:
I actually didn't notice we already have a docker-compose.yml (but it's outdated, and only the database is in a separate container...)

Simperfit:
Can we close this?

teohhanhui:
Yeah, more or less...

PR #57, opened 2016-03-24
Docker should not depend on project commands

### LINKED ISSUE #56: I should be able to build Docker environment whatever project issues
I'm currently trying to implement this project Docker configuration in my project using API Platform 2. I may have some issues on my project as it's still in development (so I could have some issues running Symfony commands like `cache:clear`).

As [Dockerfile runs `composer install`](https://github.com/api-platform/api-platform/blob/master/Dockerfile#L23), it will try to execute some Symfony commands, which will fail, which will block my Docker build.

Dockerfile should not execute any Composer or Symfony commands, to be able to create a correct Docker environment without project dependencies.

vincentchalamon:
Fix #56

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,7 +45,7 @@ WORKDIR /app
 RUN ((rm -rf var/cache/* && rm -rf var/logs/* && rm -rf var/sessions/*) || true) \
 
     # Install dependencies
-    && composer install -o && bin/console cache:warmup -e=prod \
+    && composer install -o --no-scripts \
 
     # Fixes permissions issues in non-dev mode
     && chown -R www-data . var/cache var/logs var/sessions


teohhanhui:
We could just do `composer install --no-scripts`

vincentchalamon:
I think a Dockerfile should create an environment and should not execute project commands (like `composer` or `bin/console`)

teohhanhui:
I don't see why. Ideally the Docker image should contain all dependencies so that it can be ready instantly on run.

Simperfit:
@teohhanhui I agree with @vincentchalamon the `Dockerfile` should set up the basics like nginx conf, phpini...

You can add a volume with the vendor if you want/need to, but you will have to do a `composer install` in the docker image or locally before you start to test/code.

teohhanhui:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#containers-should-be-ephemeral

vincentchalamon:
Ping @dunglas @sroze

vincentchalamon:
> with an absolute minimum of set-up and configuration

So no need to execute composer commands on it, right?

teohhanhui:
@vincentchalamon You misunderstood. It's exactly why we should do `composer install` during the image *build*, not on container *start*.

Simperfit:
@teohhanhui For me you should install composer, codeception or anything you need to be used in the image, but anything related to the project should be done in the project.

IMO, the composer files should be in a volume.

teohhanhui:
@Simperfit
> but anything related to the project should be done in the project.

The image is an image of the project. It should include the project files (source code, dependencies etc.). In practice, I use a volume for my `vendor` dir (and composer cache) and run `composer install` on container start too, but that's because of the inability to use volumes during build.

teohhanhui:
I guess it depends on what kind of use case we're trying to target here. But we need #50 for anything serious ("one container to rule them all" quickly runs into many problems).

vincentchalamon:
> By doing as much work as possible during image build, we make the container start as fast (and as simple) as possible

@teohhanhui You're right on this point. But executing a `composer install` won't slow down the container start? I mean, according to you (_we make the container start as fast as possible_), if I delete my container and start it again, it will launch a `composer install` and my container won't be available till composer finishes.

vincentchalamon:
My bad, it will be executed on the image, not in the container :wink:

vincentchalamon:
Well, in that case, I could agree with you @teohhanhui by executing a `composer install -o --no-scripts`, but I will remove the Symfony commands anyway (`cache:warmup`)

teohhanhui:
By adding the `--no-scripts` argument to the composer command, project problems should not prevent the build. :smile:

dupuchba:
@vincentchalamon I don't understand why adding the `--no-scripts`. Can you explain plz ?

vincentchalamon:
@dupuchba I got some exceptions in my Symfony project. As composer scripts will execute some Symfony commands (like `clear:cache`), I wasn't able to build my container just because of these exceptions…
Adding `--no-scripts` will disable composer scripts, so no Symfony commands anymore when building my container. I can finally build it whatever my project have some exceptions or not :)

dupuchba:
ok, got it. Thx

teohhanhui:
:+1:

gorghoa:
Hi @vincentchalamon, @teohhanhui, 

Please Consider this workflow : 

First, I create a new api platform project with `composer create-project blabla`, composer will automatically load the dependencies: 1st composer install

Second, I fire the `docker-compose up` command, then the docker build begins, executing composer install from the Dockerfile: 2nd composer install

Third, my container is up and running, thanks to docker-compose, with my project directory mounted as a volume at `/app`, therefore annihilating the  use of  the `composer install` of the docker build step since the volume uses the host machine vendors directory, not the one from the freshly created image.

Conclusion, I think the developer experience when starting a new project with a docker environment could be enhanced by stripping the composer install from the build (heck, we’ve already bear the php extensions build ;-) ).

Maybe, we should considering one Dockerfile for production (probably in a repository of his own, in order to profit of the docker hub automated builds) and one in the api-platform repo for developer use.

(Moreover, if the user public ip has overpassed the github unidentified fair use quota, the composer install will fail during the build, since the composer github auth token is not shared during the build.)

vincentchalamon:
Hi @gorghoa,
> Moreover, if the user public ip has overpassed the github unidentified fair use quota, the composer install will fail during the build, since the composer github auth token is not shared during the build.

You would be happy to know that [GitHub dropped its rate limit](https://github.com/composer/composer/issues/4884#issuecomment-195229989) some days ago :)

gorghoa:
indeed, I am happy \o/

teohhanhui:
Just came across this post by @sroze :stuck_out_tongue:

http://sroze.io/2015/12/23/immutable-docker-images/

sroze:
Yup, so as you can imagine, I definitely agree with you @teohhanhui ;)

jderusse:
Are we talking about containers for development or containers for production purpose?

dunglas:
Fixed by #87.

PR #58, opened 2016-03-25
Move Nelmio API doc to dev

vincentchalamon:


dupuchba:
:+1:

jderusse:
IMHO, documentation must be exposed publicly to consumers. It's not only a dev things.

An API is not only an endpoint, it's a documentation and SDK too

app/autoload.php
```diff
@@ -3,7 +3,7 @@
 use Composer\Autoload\ClassLoader;
 use Doctrine\Common\Annotations\AnnotationRegistry;
 
-/**
```
teohhanhui: This must be a mistake, right?

web/app_dev.php
```diff
@@ -18,7 +18,7 @@
     exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
 }
 
-/**
```
teohhanhui: Why the change?

teohhanhui:
I'm -1 for this.  Agree with @jderusse

web/app_dev.php
theofidry: in those cases (forcing a typehint with `@var` in inline code) phpdoc are unneeded, a simple comment do

teohhanhui: I'd rather we don't change things unnecessarily from symfony-standard.

In any case, this syntax is actually still wrong anyway (according to PSR-5 draft). See relevant discussion here: https://www.drupal.org/node/2305593

theofidry: I remember getting a warning from phpStorm, Scrutinizer or another tool
about that at some point which could explain why he changed that.

But in any case, thanks for looking into that anyway

vincentchalamon: This is the recommendation from StyleCI, maybe @dunglas should adapt its configuration so

vincentchalamon:
According to your comments, we keep NelmioAPIDoc as it. I'm closing this PR

PR #66, opened 2016-04-09
Fix the xml schema of the vhost.conf

n1c01a5:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

PR #73, opened 2016-04-15
Fix vhost conf

theofidry:
Revert #66 and fix indentation that lead to this error.


theofidry:
Rogue merge as it's a revert of a PR that shouldn't have been merged in the first place.

PR #79, opened 2016-05-03
Update composer.json

dunglas:
Update deps.


teohhanhui:
Don't we need to bump a few other Symfony packages too?

PR #80, opened 2016-05-03
Better default parameters

dunglas:


teohhanhui:
:+1:

dunglas:
Is is safe for Docker @sroze @teohhanhui

teohhanhui:
We should use [env_file](https://docs.docker.com/compose/compose-file/#env-file) or [environment](https://docs.docker.com/compose/compose-file/#environment) in docker-compose.yml, in conjunction with https://github.com/Incenteev/ParameterHandler

teohhanhui:
http://12factor.net/config

PR #87, opened 2016-05-23
Fix Docker setup and add Travis setup

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | todo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ services:
 
 script:
     - docker-compose build
+    - docker-compose run web composer install
     - docker-compose run web bin/console security:check
     - docker-compose run web bin/console doctrine:schema:create
     - docker-compose run web bin/console doctrine:mapping:info
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+sudo: required
+
+services:
+    - docker
+
+script:
+    - docker-compose build
+    - docker-compose run web bin/console security:check
+    - docker-compose run web bin/console doctrine:schema:create
+    - docker-compose run web bin/console doctrine:mapping:info
+    - docker-compose run web bin/console doctrine:schema:validate
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 
 script:
     - docker-compose build
-    - docker-compose run web composer install
+    - docker-compose run web composer install -o -n
     - docker-compose run web bin/console security:check
     - docker-compose run web bin/console doctrine:schema:create
     - docker-compose run web bin/console doctrine:mapping:info
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,10 +43,6 @@ WORKDIR /app
 
 # Remove cache and logs if some and fixes permissions
 RUN ((rm -rf var/cache/* && rm -rf var/logs/* && rm -rf var/sessions/*) || true) \
-
-    # Install dependencies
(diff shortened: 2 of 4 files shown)


dunglas:
I removed the `composer install` command from the `Dockerfile` because it doesn't work: vendors weren't stored in the image, probably because `/app` is a volume.

ping @sroze @vincentchalamon

vincentchalamon:
Duplicate from https://github.com/api-platform/api-platform/pull/57 ?

dunglas:
No but it replaces #57.

Differences are:

* `composer install` isn't called at all (because it doesn't work, the my previous comment)
* port mapping configuration is fixed (at least for Docker for Mac, it wasn't working previously)
* the image is now tested with Travis

vincentchalamon:
Can you detail why `composer install` command failed please? Maybe print an output of your error

dunglas:
It doesn't fail, but it is useless because /app is a mounted volume, so vendors are NOT stored in the image anyway.

vincentchalamon:
But it can be interesting to use [immutable Docker images](http://sroze.io/2015/12/23/immutable-docker-images/) as @sroze suggested, especially for testing (travis or other CI…)

dunglas:
Immutable images are very interesting for the deployment process. Here this PR has an advantage: it works.

teohhanhui:
@dunglas Your conclusion is wrong. Upon the creation of the volume, the contents are copied from the image. The only problem is when you use a host-mounted volume (or a tmpfs as discussed above for the immutable image, for that matter...)

teohhanhui:
Volumes play no part at all during Dockerfile build. They are only used during container run.

teohhanhui:
The fix is simple. Do not blindly mount the entire project directory from host, but have more fine-grained volumes instead (some host-mounted, some not).

dunglas:
We do use a host mounted volume and I consider the Docker setup (Dockerfile + Docker Compose) as whole thing for DX.

Currently it doesn't work at all. Run docker-compose up then docker run something and it will crash because vendor are missing. I'm +1 for a more fine grained solution as you suggest but this must be fixed or removed for the v1 release. I'm for merging this PR then improve the docker setup if possible.

sroze:
This Docker approach is not the ideal one to me at all. The `vendor` are
built successfully in the Docker image, but obviously have to be installed
locally too if you override the image content by mounting the `/app`
directory using your own local directory.

The port mapping configuration is something installation-specific and
shouldn't be in the Git repository to me.

teohhanhui:
@sroze:
> The port mapping configuration is something installation-specific and
shouldn't be in the Git repository to me.

It needs to be there, but can be overridden easily (e.g. by using a `docker-compose.override.yml`).

dunglas:
> This Docker approach is not the ideal one to me at all. The `vendor` are
built successfully in the Docker image, but obviously have to be installed
locally too if you override the image content by mounting the `/app`
directory using your own local directory.

It slows down the installation process (vendors are resolved and downloaded 2 times) and this isn't acceptable from a DX POV.

> The port mapping configuration is something installation-specific and
shouldn't be in the Git repository to me.

This is mandatory (see @teohhanhui comment).


Merging for now. Improvement to this are very welcome!

PR #104, opened 2016-06-20
Increase the number of books in fixtures to enable the pagination

dunglas:
It exposes more API Platform features when playing with the default installation.


dunglas:
ping @theofidry

theofidry:
Oh, I didn't think of though nice one 👍

PR #105, opened 2016-06-20
Fix Book::$title fixture generator

dunglas:
ping @theofidry


theofidry:
👍

teohhanhui:
We might want to add a custom faker to convert it to title case and strip off the period (`.`) at the end.

dunglas:
I was reluctant to add the fixture bundle in the standard edition but I thin it's definitely not the place to add things like custom faker generator.

However it can be added in the demo (the app illustrating the getting started guide) if anybody care.

teohhanhui:
I agree. We should remove all fixtures from the standard edition.

PR #106, opened 2016-06-20
Load fixtures in Travis builds

dunglas:


.travis.yml
```diff
@@ -10,3 +10,4 @@ script:
     - docker-compose run web bin/console doctrine:schema:create
     - docker-compose run web bin/console doctrine:mapping:info
     - docker-compose run web bin/console doctrine:schema:validate
+    - docker-compose run web bin/console hautelook_alice:fixtures:load
```
theofidry: `docker-compose run web bin/console hautelook_alice:doctrine:fixtures:load`

dunglas: What is the difference between the two commands? Both seems to work.

theofidry: both works, but before AliceBundle was working with Doctrine ORM only so there was no ambiguity. However since v1 you have support for ODM so `hautelook_alice:fixtures:load` has been deprecated.

PR #123, opened 2016-08-28
Remove demo files and update dependencies

dunglas:
I'll create a new repository called `api-platform/demo` including the books store app and other interesting features (authentification, groups...).


Simperfit:
LGTM 👍

PR #124, opened 2016-08-29
Fix the Docker setup

dunglas:
* Fix permissions when running Docker for Mac
* Fix Vhost configuration
* Install composer securely
* Install dependencies to speed up the next download

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,17 +25,24 @@ RUN pecl install \
     && docker-php-ext-enable --ini-name 20-apcu.ini \
         apcu
 
-# Apache & PHP configuration
+# Apache config
 RUN a2enmod rewrite
-ADD docker/apache/vhost.conf /etc/apache2/sites-enabled/default.conf
+ADD docker/apache/vhost.conf /etc/apache2/sites-available/000-default.conf
+
+# PHP config
 ADD docker/php/php.ini /usr/local/etc/php/php.ini
 
-# Install composer
+# Install Git
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         git \
     && rm -rf /var/lib/apt/lists/*
-RUN curl -sS https://getcomposer.org/installer | php \
+
+# Install composer
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
+    && php -r "if (hash_file('SHA384', 'composer-setup.php') === 'e115a8dc7871f15d853148a7fbac7da27d6c0030b848d9b3dc09e2a0388afed865e6a3d6b3c0fad45c48e2b5fc1196ae') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" \
+    && php composer-setup.php \
+    && php -r "unlink('composer-setup.php');" \
     && mv composer.phar /usr/bin/composer
 
 # Add the application
@@ -44,6 +51,9 @@ WORKDIR /app
 
 # Remove cache and logs if some and fixes permissions
 RUN ((rm -rf var/cache/* && rm -rf var/logs/* && rm -rf var/sessions/*) || true) \
+    # Install dependencies
+    && composer install --optimize-autoloader \
+    && bin/console cache:warmup -e=prod \
     # Fixes permissions issues in non-dev mode
     && chown -R www-data . var/cache var/logs var/sessions
 
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -12,7 +12,7 @@ use Symfony\Component\HttpFoundation\Request;
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -44,6 +51,9 @@ WORKDIR /app
 
 # Remove cache and logs if some and fixes permissions
 RUN ((rm -rf var/cache/* && rm -rf var/logs/* && rm -rf var/sessions/*) || true) \
+    # Install dependencies
+    && composer install --optimize-autoloader \
```
teohhanhui: Don't we need --no-scripts here? Because trying to clear cache might fail for so many reasons. What about runtime configuration options, of course most of which are absent at build time?

For the same reason I'm not sure we should do `cache:warmup` here.

web/app_dev.php
```diff
@@ -12,7 +12,7 @@
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1']) || php_sapi_name() === 'cli-server')
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1', '172.17.0.1']) || php_sapi_name() === 'cli-server')
```
teohhanhui: I'm against this change. This is the `docker0` host IP for the default Docker bridge network. The user can customize it, and there is even no guarantee that it'll remain this way (apparently it has changed before - docker/docker#17744).

More importantly, I do not believe that the app should be made in such a way that it caters specifically to Docker (or any such environment).

Dockerfile
dunglas: It works with default scripts. If someone add something breaking it, he an add the --no-scripts flags by itself. WDYT?

web/app_dev.php
dunglas: Not having it is a major DX issue. It makes it almost impossible to use Docker for dev.

It's a private IP, AFAIK it should not cause security issue and the user can still edit the file to fit his needs.

If Docker changes the IP aain, we will update the file. If the user change it, he must update his file himself.

teohhanhui: We've always made do with local modifications of the file (commenting this check out).

Coupled with `git update-index --skip-worktree web/app_dev.php` it works perfectly fine.

Dockerfile
teohhanhui: Clearing the cache is actually plenty quick. It can be done at container startup.

teohhanhui: We actually do another `composer install` in the startup script (because of Docker volumes). It completes almost instantly.

web/app_dev.php
dunglas: I'll push robuster alternative implem.

dunglas: What do you think of this one?

teohhanhui: Better :)

web/app_dev.php
```diff
@@ -10,9 +10,14 @@
...
 // Feel free to remove this, extend it, or make something more sophisticated.
+$whiteListedAddresses = ['127.0.0.1', 'fe80::1', '::1'];
+if ($dockerBridgeIp = getenv('DOCKER_BRIDGE_IP')) {
+    $whiteListedAddresses[] = $dockerBridgeIp;
+}
+
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', 'fe80::1', '::1']) || php_sapi_name() === 'cli-server')
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], $whiteListedAddresses) || php_sapi_name() === 'cli-server')
```
teohhanhui: Taking a closer look, I see that this is still not a tenable solution, because `isset($_SERVER['HTTP_X_FORWARDED_FOR'])` will be true for a nginx + php-fpm setup.

We should ideally not try to be smart here and just let the user decide for themselves on how to overcome this. A note in the documentation is of course good.

teohhanhui: In other words, we cannot cater for all the different scenarios. It's worse to have to maintain specific workarounds than just leaving it to the user, like what symfony-standard does.

dunglas: Well, we're only using Apache right now. `symfony-standard` doesn't have Docker support at all. If we add Docker support (and we should because it improves the whole DX, like Homestead do for Laravel), we must propose a flawless **default** experience for the user. 

Ideally it should be as easy as "Download and unzip api-platform.zip then run `docker-compose up`" (it's something like that I'm actually writing in the new "getting started" guide).

FPM + Nginx is not yet an officially supported setup (but I'm 👍 to move to this setup if we find solutions to preserve DX). API Platform v2 will likely be released with Apache support only. I propose merge this PR for now and find a solution dedicated to nginx when we'll move to it (probably v2.1). WDYT?

teohhanhui: Yes, I think this is a reasonable middle ground for now. But of course we have to be conscious of how much we want to deviate from symfony-standard.

PR #139, opened 2016-09-27
Change inline var PHPDoc syntax

teohhanhui:
Following symfony/symfony-standard#990

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -11,9 +11,7 @@ use Symfony\Component\Debug\Debug;
 
 set_time_limit(0);
 
-/**
- * @var Composer\Autoload\ClassLoader $loader
- */
+/** @var \Composer\Autoload\ClassLoader $loader */
 $loader = require __DIR__.'/../app/autoload.php';
 
 $input = new ArgvInput();
--- a/web/app.php
+++ b/web/app.php
@@ -2,9 +2,7 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
-/*
- * @var Composer\Autoload\ClassLoader
- */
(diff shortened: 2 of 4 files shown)


Simperfit:
👍

bin/console
```diff
@@ -11,9 +11,7 @@ use Symfony\Component\Debug\Debug;
 
 set_time_limit(0);
 
-/**
- * @var Composer\Autoload\ClassLoader $loader
- */
+/** @var \Composer\Autoload\ClassLoader $loader */
```
amenophis: Use a ```use``` at the top of the class, no ?

web/app.php
```diff
@@ -2,9 +2,7 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
-/*
- * @var Composer\Autoload\ClassLoader
- */
+/** @var \Composer\Autoload\ClassLoader $loader */
```
amenophis: same below

web/app_dev.php
```diff
@@ -23,9 +23,7 @@
     exit('You are not allowed to access this file. Check '.basename(__FILE__).' for more information.');
 }
 
-/**
- * @var Composer\Autoload\ClassLoader
- */
+/** @var \Composer\Autoload\ClassLoader $loader */
```
amenophis: same as below

bin/console
teohhanhui: I'd prefer that we do not deviate from `symfony/symfony-standard` unnecessarily.

dunglas:
Thanks @teohhanhui

PR #141, opened 2016-09-28
Move Action loading after api

amenophis:
If you add a custom itemOperation, the route will be before in the RouteCollection, and will be used for generating the @id attribute.

This patch move the custom routes after the api generated.

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -1,7 +1,7 @@
-app:
-    resource: '@AppBundle/Action/'
-    type:     'annotation'
-
 api:
     resource: '.'
     type:     'api_platform'
+
+app:
+    resource: '@AppBundle/Action/'
+    type:     'annotation


dunglas:
Thank you @amenophis

teohhanhui:
I don't believe that this is the right fix. If we're using the wrong route for IRI generation, we should fix that instead.

dunglas:
@teohhanhui I'm not sure to understand what you mean. For now we are using the first GET route registered for a given resource to generate the IRI. It means than before the patch of @amenophis, when you add a custom action, its route becomes the route used to generate the IRI. It doesn't look like a good default.

teohhanhui:
We're talking about api-platform/core#689 right?

From your comment there: https://github.com/api-platform/core/issues/689#issuecomment-246192456

> But there is something weird here, if you register the typical get operation first, it should always be used to generate the URL. If it's not the case, it's a bug and we must fix it.

So indeed this PR is merely a workaround of that bug, hence my objection.

dunglas:
@teohhanhui yes. But both are not incompatible. It would be nice to have a new option to set the route to use for IRI generation. During the meantime (and even after) this patch doesn't hurt, it improves the DX easily.

teohhanhui:
@dunglas I thought in #689 we were talking about the order in which the *item operations* are declared for `ResourceMetadata`, not the order of Symfony routes in the `RouteCollection`?

dunglas:
IIRC it's the order of Symfony routes that matters. IMO It's not a bug, but the DX will definitely be improved if we have a flag to mark the main route instead.

PR #150, opened 2016-10-13
Add scripts for swagger-ui

soyuka:
see https://github.com/api-platform/core/pull/768

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
         "incenteev/composer-parameter-handler": "^2.0",
         "dunglas/action-bundle": "^0.3",
         "nelmio/cors-bundle": "^1.4",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -78,5 +78,16 @@
         "branch-alias": {
             "dev-master": "2.0-dev"
         }
+    },
+    "scripts": {
```
teohhanhui: We cannot have duplicate `scripts` key. There's already an existing section above with the Symfony scripts.

soyuka:
Done

soyuka:
This has to be merged in combination with api-platform/core#768

Simperfit:
Could you check why the test are not green :p ?

soyuka:
May it has to be with the fact that  api-platform/core#768 isn't available yet?

PR #162, opened 2016-10-25
Feature/docker update

jockos:
Update docker configuration to use nginx.

Currently trying to setup a config file for production environment.


jockos:
ping @dunglas

dunglas:
@sroze @teohhanhui can you take a look at this one?

As the current Apache setup is working pretty well, I propose to wait after the release of the 2.0 to merge this PR (it will be included in API Platform 2.1).

Dockerfile
```diff
@@ -1,43 +1,53 @@
...
+ADD ./docker/php/php.ini /usr/local/etc/php/php.ini
 
 # Install Git
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         git \
     && rm -rf /var/lib/apt/lists/*
 
+# Add the application
+ADD ./ /var/www/app
```
teohhanhui: May I suggest `/srv/app`?

teohhanhui: And same as above, COPY instead of ADD.

Dockerfile
```diff
@@ -1,43 +1,53 @@
...
+ADD ./docker/nginx/conf/nginx.conf /etc/nginx/conf/nginx.conf
```
teohhanhui: What is nginx configuration doing in the php-fpm container? Please enlighten me...

docker-compose.yml
```diff
@@ -1,16 +1,45 @@
...
-    image: mysql
-    environment:
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
+    nginx:
+        container_name: nginx
```
teohhanhui: There is no need to use hardcoded container_name. DNS resolving works just fine with Docker networking.

docker-compose.yml
```diff
@@ -1,16 +1,45 @@
...
+        image: nginx:mainline-alpine
+        restart: always
+        dns_search: nginx
```
teohhanhui: This is also unnecessary.

docker-compose.yml
```diff
@@ -1,16 +1,45 @@
...
-    environment:
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
+    nginx:
+        container_name: nginx
+        image: nginx:mainline-alpine
```
teohhanhui: We should use proper versioning. I suggest something like `nginx:1.11-alpine`.

docker-compose.yml
```diff
@@ -1,16 +1,45 @@
...
+        restart: always
+        dns_search: php
+        build: ./
+        volumes:
+            - ./:/var/www/app
+        depends_on:
+            - database
+    database:
+        container_name: mysql
+        image: mysql:latest
```
teohhanhui: Same here. Versioning is important so perhaps `mysql:5.7`.

docker-compose.yml
```diff
@@ -1,16 +1,45 @@
...
+        restart: always
+        volumes:
+            - data:/var/lib/mysql
+        ports:
+            - "3306:3306"
+        expose:
```
teohhanhui: `expose` is redundant when same port number binds to host using `ports`.

docker-compose.yml
```diff
@@ -1,16 +1,45 @@
...
+            - "3306"
+        environment:
+            MYSQL_ROOT_PASSWORD: changeme
+            MYSQL_DATABASE: api_platform
+            MYSQL_USER: api_platform
+            MYSQL_PASSWORD: api_platform
+        dns_search: database
+volumes:
+    data:
```
teohhanhui: I suggest `database-data` (or nicer, `db-data` if we rename the mysql service to `db`).

docker/php/run.sh
```diff
@@ -3,6 +3,5 @@ set -xe
 
 # Detect the host IP
 export DOCKER_BRIDGE_IP=$(ip ro | grep default | cut -d' ' -f 3)
-
-# Start Apache with the right permissions
-/app/docker/apache/start_safe_perms -DFOREGROUND
+composer install --optimize-autoloader --no-scripts
```
teohhanhui: There is no point in running this here, as the same line is already in the Dockerfile. If we run without `--no-scripts` here, then it'd be desired.

But still I'd like to see a more comprehensive composer command with `--prefer-dist`, at least.

docker/php/run.sh
```diff
@@ -3,6 +3,5 @@ set -xe
...
+php-fpm
```
teohhanhui: I never noticed this, but this is just asking for trouble. Basic Docker fail :stuck_out_tongue:

`exec php-fpm`

Dockerfile
```diff
@@ -1,43 +1,53 @@
...
-        libicu52 \
-        zlib1g \
+       $buildDeps \
+       libicu52 \
+       zlib1g \
     && rm -rf /var/lib/apt/lists/* \
     && docker-php-ext-install \
         intl \
         mbstring \
+        pdo \
```
jockos: Can you tell me where ? 
Because I can not find where it is installed, tried to run the config without that line and it is not working.

jockos:
@teohhanhui I've made some changes according to your comments.

Dockerfile
teohhanhui: http://php.net/manual/en/pdo.installation.php

> PDO and the PDO_SQLITE driver is enabled by default as of PHP 5.1.0. You may need to enable the PDO driver for your database of choice; consult the documentation for database-specific PDO drivers to find out more about that.

Dockerfile
```diff
@@ -1,60 +1,58 @@
...
+	&& apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		$buildDeps \
+		libicu52 \
+		zlib1g \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& docker-php-ext-install \
+		intl \
+		mbstring \
+		pdo \
```
teohhanhui: As mentioned, `pdo` is unnecessary here.

Dockerfile
```diff
@@ -1,60 +1,58 @@
...
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-        git \
-    && rm -rf /var/lib/apt/lists/*
+	&& apt-get install -y --no-install-recommends \
+		git \
+	&& rm -rf /var/lib/apt/lists/*
+
+# Add the application
+COPY ./ /srv/app
```
teohhanhui: It'd be good if we can avoid blindly copying the entire directory. Hmm...

docker-compose.yml
```diff
@@ -1,16 +1,37 @@
...
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
+    nginx:
+        image: nginx:1.11-alpine
+        restart: always
+        dns_search: nginx
```
teohhanhui: Why is `dns_search` necessary?

docker-compose.yml
```diff
@@ -1,16 +1,37 @@
...
-database:
-    image: mysql
-    environment:
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
+    nginx:
```
teohhanhui: It'd be good to sort the keys (services) alphabetically.

docker-compose.yml
```diff
@@ -1,16 +1,37 @@
...
+        image: nginx:1.11-alpine
```
teohhanhui: It'd be good to sort the keys (options) alphabetically.

docker-compose.yml
```diff
@@ -1,16 +1,37 @@
...
+            - "3306:3306"
+        environment:
+            MYSQL_ROOT_PASSWORD: changeme
+            MYSQL_DATABASE: api_platform
+            MYSQL_USER: api_platform
+            MYSQL_PASSWORD: api_platform
+        dns_search: db
+volumes:
+    db-data:
+        ~
```
teohhanhui: `db-data: {}`

docker/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,53 @@
...
+        internal;
+    }
+
+    # return 404 for all other php files not matching the front controller
+    # this prevents access to other php files you don't want to be accessible.
+    location ~ \.php$ {
+      return 404;
+    }
+
+    error_log /var/log/nginx/project_error.log;
```
teohhanhui: Perhaps:

app_error.log
app_access.log

docker/nginx/conf/nginx.conf
```diff
@@ -0,0 +1,39 @@
+worker_processes 1;
```
teohhanhui: I suggest removing this file.

Dockerfile
jockos: I agree but what do you suggest to do instead ? Copying folder by folder seems a bit redundant...
Maybe we can only copy the application files and folder and exclude the tests & docker folders.

docker/nginx/conf.d/default.conf
jockos: Would it be interesting to mount a volume in /var/log/nginx to get the logs from nginx ?

Dockerfile
amenophis: You can use .dockerignore file, it is like .gitignore but for docker ;)
https://docs.docker.com/engine/reference/builder/#/dockerignore-file

docker/nginx/conf.d/default.conf
amenophis: Why not use docker output ? It is the normal behavior for the nginx offcial image

dunglas: I agree, the standard output is the normal way to deal with logs on Docker (and many tools require that this behavior works).

Dockerfile
jockos: Ok thank you :)
Do we need to add features, test and small files like README, composer.lock in the container ?
If it is for production it doesn't seems necessary to me if we want very light container.

teohhanhui: I think we should copy the folder by folder. `.dockerignore` is useful to exclude things that we never want to be included in the build context (e.g. Dockerfile, docker-compose.yml, .git, .gitignore).

docker-compose.yml
teohhanhui: @jockos You have not addressed this comment. :smile:

jockos: Well, in that case it's not necessary ^^' so I removed it in my last commit :)

docker-compose.yml
```diff
@@ -1,16 +1,35 @@
...
-    image: mysql
-    environment:
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
+    db:
+        dns_search: db
```
teohhanhui: Is this necessary? I have not had the need to use it in our project. But please enlighten me...

docker-compose.yml
```diff
@@ -1,16 +1,35 @@
...
+            - "443:443"
+        restart: always
+        volumes:
+            - ./:/srv/app
+            - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
+    php:
+        build: ./
+        depends_on:
+            - db
+        dns_search: php
```
teohhanhui: And this too...

docker-compose.yml
jockos: We use it in the parameter.yml for the database_host.

docker-compose.yml
jockos: Php is used in the nginx config file line 12 (fastcgi_pass php:9000;)

docker-compose.yml
teohhanhui: The service name is automatically aliased in the built-in DNS for the Docker network. It has been working for me since a long time ago (I forgot when it was added), so make sure you're using a recent version of Docker and Docker Compose...

jockos: Ok that's good to know, I wasn't aware of that, thanks. :)

.dockerignore
```diff
@@ -1,26 +1,14 @@
...
-*.sublime-project
-*.sublime-workspace
-._*
-.dockerignore
-.editorconfig
-.git/
-.gitignore
-.gitmodules
-.travis.yml
+*
```
teohhanhui: Please don't ignore everything... :cry:

We do not want surprises for the user. Let's just use a blacklist. (I don't think you need to change much from the existing .dockerignore)

Dockerfile
```diff
@@ -1,60 +1,59 @@
...
+	&& composer global require "hirak/prestissimo:^0.3"
+
+# Remove cache and logs if some and fixes permissions
+RUN ((rm -rf var/cache/* && rm -rf var/logs/* && rm -rf var/sessions/*) || true) \
+	# Install dependencies
+	&& composer install --optimize-autoloader --no-scripts \
+	# Fixes permissions issues in non-dev mode
+	&& chown -R www-data . var/cache var/logs var/sessions
+
+CMD ["/srv/app/docker/php/run.sh"]
```
teohhanhui: Can we copy what we need into the container instead of having the ./docker directory copied into the image too? (Totally unnecessary...)

Dockerfile
```diff
@@ -1,60 +1,59 @@
...
-        git \
-    && rm -rf /var/lib/apt/lists/*
+	&& apt-get install -y --no-install-recommends \
+		git \
+	&& rm -rf /var/lib/apt/lists/*
 
 # Add the application
-ADD . /app
-WORKDIR /app
+COPY ./ /srv/app
```
teohhanhui: Instead of copying everything, we can be more selective:

```Dockerfile
# after "Install composer" section

WORKDIR /srv/app

COPY composer.json ./
COPY composer.lock ./

RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest

COPY app app/
COPY bin/console bin/
COPY src src/
COPY web web/

COPY docker/php/start.sh /
COPY docker/php/update-acl.sh /

CMD ["/start.sh"]
```

Dockerfile
```diff
@@ -1,60 +1,59 @@
...
+RUN php -r "copy('https://composer.github.io/installer.sig', 'installer.sig');" \
+	&& php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
+	&& php -r "if (hash_file('SHA384', 'composer-setup.php') === trim(file_get_contents('installer.sig'))) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" \
+	&& php composer-setup.php \
+	&& php -r "unlink('composer-setup.php');" \
+	&& php -r "unlink('installer.sig');" \
+	&& mv composer.phar /usr/bin/composer \
+	&& composer global require "hirak/prestissimo:^0.3"
+
+# Remove cache and logs if some and fixes permissions
```
teohhanhui: We don't need this section if we make sure the directories are mounted as volumes (even in production, yes - for performance reasons).

Dockerfile
teohhanhui: Contents of `update-acl.sh`:

```bash
#!/bin/bash
set -e

setfacl -R -m u:www-data:rwX -m u:`whoami`:rwX var
setfacl -dR -m u:www-data:rwX -m u:`whoami`:rwX var
```

docker-compose.yml
```diff
@@ -1,18 +1,33 @@
...
+        restart: always
+        volumes:
+            - ./:/srv/app
+            - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
+    php:
+        build: ./
+        depends_on:
+            - db
+        volumes:
+            - ./:/srv/app
```
teohhanhui: @api-platform/core-team We should have a docker-compose.override.yml file. But the problem is that we need to have different templates (.dist files) for different environments, which means the initial setup needs a preparatory step.

docker/start.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/sh
```
teohhanhui: Please use this as docker/php/start.sh and remove docker/php/run.sh

docker-compose.yml
Simperfit: I really don't know what to think about this  : 
- Do we really want to support people using docker in prod, or do we made this like a bootstrap ?
- Why don't we build an image and use it ?

dunglas:
Hi @jockos, @teohhanhui,

What do you thin about using [PHP PM](https://github.com/php-pm/php-pm) instead of PHP FPM in this Docker setup? It increases the performance of API Platform 10 times (according to my tests).

Dockerfile
jockos: Setfacl sends "Operation not supporter and I think that it is because docker uses AUFS and it doesn't support setfacl.

But this is working : 
`usermod -u 1000 www-data`
`chown -R www-data:www-data /var/www/app/cache`
`chown -R www-data:www-data /var/www/app/logsr`

What do you think ?

.travis.yml
```diff
@@ -5,5 +5,5 @@ services:
 
 script:
     - docker-compose build
-    - docker-compose run --rm web composer install -o -n
-    - docker-compose run --rm web bin/console security:check
+    - docker-compose run php composer install -o -n
```
vincentchalamon: Should be better to keep `--rm` option, or use `exec` instead of `run`, it will prevent useless containers

Dockerfile
vincentchalamon: Why not simply use `apipltfrm/demo` in `docker-compose.yml` file instead? The image should be the same, isn't it?

docker-compose.yml
teohhanhui: To me the one of the appeals of Docker is that I can use (almost) the same setup for both development and production. But I understand that there's no one-size-fits-all solution, and for a "Standard Edition" like what we have here, we may want to target development only.

.travis.yml
teohhanhui: Yes, we can do:

```sh
docker-compose up -d php
docker-compose exec php composer install -o -n
...
```

Dockerfile
teohhanhui: @vincentchalamon:

> Why not simply use apipltfrm/demo in docker-compose.yml file instead? The image should be the same, isn't it?

I don't think they should be the same.

teohhanhui:
> What do you think about using PHP PM instead of PHP FPM in this Docker setup? It increases the performance of API Platform 10 times (according to my tests).

I'm -1 for using PHP-PM by default. It only presents more gotchas and head-scratching. If the user decides to switch to PHP-PM, he/she should be fully aware of the decision and what to look out for.

Dockerfile
jockos: @teohhanhui what do you mean by "made into volumes" ?
Do you want a volume container or just the Volume instruction in the Dockerfile ?

webda2l: To avoid copying data inner this docker, the data-only containers pattern use, could be nicer no? http://stackoverflow.com/a/36187864/520114

And, by the way, more globally speaking about this Dockerfile, to continue with the already nginx alpine use, the php:7-fpm-alpine use instead php:7.0-fpm will be nicer.

teohhanhui: Data containers is an outdated pattern. Do not use that.

Yes, I meant adding VOLUME instructions.

teohhanhui: @webda2l If we use the alpine variant, we might run into problems with third-party tools (e.g. blackfire - but probably they've fixed that after my report last time). Compatibility could be an issue...

jockos: @teohhanhui I think I misunderstood the use of the VOLUME instruction...
I tried to replace COPY by VOLUME but it's not working.
What the goal of doing that ?

Dockerfile
```diff
@@ -1,60 +1,66 @@
...
+	&& php -r "if (hash_file('SHA384', 'composer-setup.php') === trim(file_get_contents('installer.sig'))) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" \
+	&& php composer-setup.php \
+	&& php -r "unlink('composer-setup.php');" \
+	&& php -r "unlink('installer.sig');" \
+	&& mv composer.phar /usr/bin/composer \
+	&& composer global require "hirak/prestissimo:^0.3"
+
+WORKDIR /srv/app
+
+VOLUME composer.json ./
```
teohhanhui: This is wrong... :joy:

```Dockerfile
COPY composer.json ./
COPY composer.lock ./
```

Dockerfile
```diff
@@ -1,60 +1,66 @@
...
+VOLUME composer.lock ./
+
+RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest
+
+VOLUME app app/
```
teohhanhui: Same here...

```Dockerfile
COPY app app/
COPY bin/console bin/
COPY src src/
COPY web web/
```

Dockerfile
```diff
@@ -1,60 +1,66 @@
...
+VOLUME bin/console bin/
+VOLUME src src/
+VOLUME web web/
+
+VOLUME docker/php/start.sh /
```
teohhanhui: And here...

```Dockerfile
COPY docker/php/start.sh /
COPY docker/php/update-acl.sh /
```

Dockerfile
```diff
@@ -1,60 +1,66 @@
...
+VOLUME docker/php/update-acl.sh /
+
+RUN chmod +x /start.sh
+RUN chmod +x /update-acl.sh
+
```
teohhanhui: We add the VOLUME instructions last (to preserve changes from any previous steps):

```Dockerfile
VOLUME /srv/app/var
VOLUME /srv/app/var/cache
VOLUME /srv/app/var/logs
VOLUME /srv/app/var/sessions
VOLUME /srv/app/vendor
```

teohhanhui: As an alternative, we can add all these volumes in the docker-compose.yml

I'm not sure which is the better / more correct way.

dunglas:
@jockos can you rebase please? It's ready to be merged?

docker-compose.yml
```diff
@@ -1,18 +1,38 @@
...
+        restart: always
+        volumes:
+            - db-data:/var/lib/mysql
+    nginx:
+        image: nginx:1.11-alpine
+        depends_on:
+            - php
+        ports:
+            - "80:80"
+            - "443:443"
```
dunglas: Why do we expose this port? IIUC SSL isn't configured.

jockos:
@dunglas Travis failed because of a port already in use... but it works fine when I test it locally.

Dockerfile
```diff
@@ -1,60 +1,64 @@
...
+	&& php -r "unlink('composer-setup.php');" \
+	&& php -r "unlink('installer.sig');" \
+	&& mv composer.phar /usr/bin/composer \
+	&& composer global require "hirak/prestissimo:^0.3"
+
+WORKDIR /srv/app
+
+COPY composer.json ./
+COPY composer.lock ./
+
```
teohhanhui: We should do a `composer install` here:

```Dockerfile
RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest \
	&& composer clear-cache
```

Dockerfile
```diff
@@ -1,60 +1,64 @@
...
-
-CMD ["/app/docker/start.sh"]
+RUN php -r "copy('https://composer.github.io/installer.sig', 'installer.sig');" \
+	&& php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
+	&& php -r "if (hash_file('SHA384', 'composer-setup.php') === trim(file_get_contents('installer.sig'))) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" \
+	&& php composer-setup.php \
+	&& php -r "unlink('composer-setup.php');" \
+	&& php -r "unlink('installer.sig');" \
+	&& mv composer.phar /usr/bin/composer \
+	&& composer global require "hirak/prestissimo:^0.3"
```
teohhanhui: We should clear the composer cache at the end:

```Dockerfile
RUN ... \
	&& ... \
	&& composer clear-cache
```

docker/php/update-acl.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
teohhanhui: This is a very wrong approach. :disappointed:

dunglas:
We also need to tweak the `php.ini` file (http://symfony.com/doc/current/performance.html):

```ini
opcache.max_accelerated_files = 20000
realpath_cache_size=4096K
realpath_cache_ttl=600
```

docker/php/update-acl.sh
jockos: I tried setfacl but it's not handled (as I said in a previous comment) so that's why I used this instead. 
Please feel free to tell me if you have a better approach :)

docker-compose.yml
```diff
@@ -1,18 +1,37 @@
...
+services:
+    db:
+        environment:
+            MYSQL_DATABASE: api_platform
+            MYSQL_PASSWORD: api_platform
+            MYSQL_ROOT_PASSWORD: changeme
+            MYSQL_USER: api_platform
+        image: mysql:5.7
+        ports:
+            - "3306:3306"
```
dunglas: Is it necessary to publicly expose this port? I prefer to let the use doing this kind security related changes by himself.

docker-compose.yml
```diff
@@ -1,18 +1,37 @@
...
+        image: nginx:1.11-alpine
+        depends_on:
+            - php
+        ports:
+            - "80:80"
+        restart: always
+        volumes:
+            - ./:/srv/app
+            - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
+    php:
```
dunglas: Why do you use `restart: always` for every containers except this one?

docker/php/update-acl.sh
dunglas: I don't find your previous comment. What is the problem exactly? You need to install the `acl` package thru APT to be able to execute the `setfacl` command.

dunglas: Double thinking about this, I'm unsure that Docker support ACL. Anyway, we cannot rely on hardcoding the 1000 uid, see https://github.com/api-platform/api-platform/blob/master/docker/apache/start_safe_perms#L8-L24

Simperfit: Docker does not support ACL

jockos: So what should we do about it ?

sroze:
Let me have a look tomorrow

docker/php/update-acl.sh
teohhanhui: It's not Docker that does not support ACL, but the underlying filesystem. If you mean AUFS, yes, it does not support ACL. Mounting volumes works around the problem (that's what I do).

But of course on Mac there's an additional hurdle in the form of osxfs (the filesystem provided by Docker for Mac).

docker-compose.yml
teohhanhui: I don't think we should include `restart: always`. That should be something for the user to decide.

Dockerfile
teohhanhui: We do composer install before copying over project files in order to properly utilize caching of Docker image layers. Otherwise any single change in the project files would cause a rebuild.

Dockerfile
```diff
@@ -1,60 +1,67 @@
...
-ADD . /app
-WORKDIR /app
+COPY app app/
+COPY bin bin/
+COPY docker/composer.sh ./
+COPY src src/
+COPY web web/
+
+COPY docker/php/start.sh /
+COPY docker/php/update-acl.sh /
```
dunglas: You need to `chmod +x` all executable files to avoid problems on Windows.

Dockerfile
```diff
@@ -1,60 +1,67 @@
...
-    # Fixes permissions issues in non-dev mode
-    && chown -R www-data . var/cache var/logs var/sessions
-
-CMD ["/app/docker/start.sh"]
+	(rm -rf var || true) \
+	&& mkdir -p var/cache var/logs var/sessions \
+	&& composer install --prefer-dist --no-dev --optimize-autoloader --classmap-authoritative --no-scripts --no-progress --no-suggest \
+	&& composer clear-cache
+
+RUN chmod +x /start.sh
```
dunglas: It can be merged in one line, dans I'll move this line just after the `COPY` instruction for those files.

Dockerfile
```diff
@@ -1,60 +1,67 @@
...
+COPY composer.lock ./
 
-# Add the application
-ADD . /app
-WORKDIR /app
+COPY app app/
+COPY bin bin/
+COPY docker/composer.sh ./
+COPY src src/
+COPY web web/
```
soyuka: Can I ask why we don't just `ADD . /app` anymore?

app/config/parameters.yml.dist
```diff
@@ -2,7 +2,7 @@
 # Set parameters here that may be different on each deployment target of the app, e.g. development, staging, production.
 # http://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
 parameters:
-    database_host:     database
+    database_host:     db
```
soyuka: database was fine no?

docker-compose.yml
```diff
@@ -1,18 +1,36 @@
...
+        restart: always
+        volumes:
+            - ./var:/srv/app/var
+            - ./var/cache:/srv/app/var/cache
+            - ./var/logs:/srv/app/var/logs
+            - ./var/sessions:/srv/app/var/sessions
+            - ./vendor:/srv/app/vendor
+            - ./:/srv/app
+volumes:
+    db-data: {}
```
soyuka: Is this necessary?

docker/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,53 @@
...
+
+    # return 404 for all other php files not matching the front controller
+    # this prevents access to other php files you don't want to be accessible.
+    location ~ \.php$ {
+      return 404;
+    }
+
+    error_log /var/log/nginx/app_error.log;
+    access_log /var/log/nginx/app_access.log;
+}
```
soyuka: I assume this configuration was copy/pasted from somewhere? Mind mentioning from where?

docker/php/update-acl.sh
soyuka: I mentioned this in https://github.com/api-platform/api-platform/pull/218#issuecomment-275159435 IMO just having `php-fpm` run as `www-data` should be sufficient! This should be possible as [`php-fpm`](https://github.com/docker-library/php/blob/master/7.0/fpm/Dockerfile#L172) exposes 9000!

Then `nginx`, that runs as `root` just forwards ports.

Dockerfile
```diff
@@ -1,60 +1,67 @@
...
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends \
-        git \
-    && rm -rf /var/lib/apt/lists/*
+	&& apt-get install -y --no-install-recommends \
+		git \
+	&& rm -rf /var/lib/apt/lists/*
+
+WORKDIR /srv/app
+
```
soyuka: `USER www-data`

https://github.com/api-platform/api-platform/pull/162/files#r97947263

docker/nginx/conf.d/default.conf
jockos: Yes, from Symfony official [page](http://symfony.com/doc/current/setup/web_server_configuration.html#nginx)

soyuka:
Okay, I get back what I said about `www-data`. I experimented a lot to find the best solution for permissions and I've come with this solution: https://github.com/soyuka/api-platform-docker-test. 

@jockos wdyt?

In this pull request some things should be changed:

1. The `docker-compose` volume for nginx should be a `volume_from` referencing the `php` image
2. COPY is indeed better than ADD for what we need. Still, we don't need to copy the files that are used through the volume. This also reduces the number of volumes to `1`.
3. IMHO, we should avoid using paths like `/srv`, but instead use `/api-platform`, it's cleaner
4. The same goes for the `start.sh` script, `docker_endpoint.sh` or `docker_cmd.sh` is better
5. If we use alpine for nginx, we should use it for php-fpm too
6. composer should not run as root, instead we should use `gsudo` (on debian) or `su-exec` (on alpine) (this means creating a `www-data` user anyway)
7. With this line, we should be able to avoid any script that tries to fix the permissions:

```
docker-compose run --rm php chown -R $(id -u):$(id -g) .
```

dunglas:
> If we use alpine for nginx, we should use it for php-fpm too

Alpine isn't compatible with some extensions yet including Blackfire (but a version supporting Alpine will be released soon). Maybe can we switch to Alpine but add a comment to make it explicit that some extensions aren't compatible with it and some instructions to switch to Debian.

soyuka:
> Alpine isn't compatible with some extensions yet including Blackfire (but a version supporting Alpine will be released soon). Maybe can we switch to Alpine anyway but add a comment to make it explicit that some extensions aren't compatible with it and some instructions to switch to Debian.

We can also use Debian for nginx and Debian for php-fpm. It's just to avoid pulling to many files as docker really eats lot of space. Your call!

jockos:
> The docker-compose volume for nginx should be a volume_from referencing the php image

I don't think we should use `volume_from`, it is now deprecated in docker-compose file V3.

soyuka:
In v3, which hasn't been released yet (or has it?). We're setting `version: '2'`, therefore I think we can use it safely.

teohhanhui:
`volumes_from` is implicit. We should instead name our volumes (unless, of course, it's bind-mounted from host) and mount each of them explicitly.

jockos:
@soyuka @teohhanhui Can you check the changes I have made ?

docker-compose.yml
```diff
@@ -1,18 +1,35 @@
...
-database:
-    image: mysql
-    environment:
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
+    db:
```
Simperfit: Are we ok with mysql ? can't we use postgres?

app/config/parameters.yml.dist
Simperfit: db is fine too :p

Dockerfile
Simperfit: Maybe to just add what we need to and to avoid adding .git ?

Dockerfile
```diff
@@ -1,63 +1,59 @@
...
-    " \
-    && apt-get update \
-    && apt-get install -y --no-install-recommends \
-        $buildDeps \
-        libicu52 \
-        zlib1g \
-    && rm -rf /var/lib/apt/lists/* \
-    && docker-php-ext-install \
-        intl \
+RUN apk add --no-cache --virtual build-dependencies \
```
Simperfit: did you run the test within the alpine?

jockos: What do you mean ?

docker-compose.yml
dunglas: I prefer Postgres myself but the standard installation must stay as close as possible as most devs already know. So I'm for keeping MySQL for now.

Dockerfile
soyuka: We have a `.dockerignore` for this.

Dockerfile
```diff
@@ -1,63 +1,59 @@
...
-        zlib1g-dev \
-    " \
-    && apt-get update \
-    && apt-get install -y --no-install-recommends \
-        $buildDeps \
-        libicu52 \
-        zlib1g \
-    && rm -rf /var/lib/apt/lists/* \
-    && docker-php-ext-install \
+RUN apk add --no-cache --virtual build-dependencies \
```
teohhanhui: Please use this snippet instead:

```Dockerfile
RUN apk add --no-cache --virtual .run-deps \
		git \
		icu-libs \
		su-exec \
		zlib

ENV APCU_VERSION 5.1.8

RUN set -xe \
	&& apk add --no-cache --virtual .build-deps \
		$PHPIZE_DEPS \
		icu-dev \
		zlib-dev \
	&& docker-php-ext-install \
		intl \
		mbstring \
		pdo_mysql \
		zip \
	&& pecl install \
		apcu-${APCU_VERSION} \
	&& docker-php-ext-enable --ini-name 20-apcu.ini apcu \
	&& docker-php-ext-enable --ini-name 05-opcache.ini opcache \
	&& apk del .build-deps
```

Dockerfile
teohhanhui: `.dockerignore` is not meant to be abused like that. We should only add what we need to the image. It should be explicit. :)

Dockerfile
```diff
@@ -1,63 +1,59 @@
...
+
+COPY composer.json ./
+COPY composer.lock ./
 
-# Add the application
-ADD . /app
-WORKDIR /app
+COPY app app/
+COPY bin bin/
+COPY docker/composer.sh ./
```
teohhanhui: This is wrong. We wouldn't want to pollute the working directory. It should be changed to:

```Dockerfile
COPY docker/composer.sh /usr/local/bin/docker-app-install-composer
```

Dockerfile
```diff
@@ -1,63 +1,59 @@
...
+COPY src src/
+COPY web web/
 
-# Fix permissions (useful if the host is Windows)
-RUN chmod +x docker/composer.sh docker/start.sh docker/apache/start_safe_perms
+COPY docker/php/start.sh /
```
teohhanhui: Following the new convention (e.g. https://github.com/docker-library/php/blob/fe1ff7ee08861f85cd1c61b54c53550435f4b915/7.1/fpm/alpine/Dockerfile#L140), I think this should be changed to:

```Dockerfile
COPY docker/php/start.sh /usr/local/bin/docker-app-start
```

Dockerfile
```diff
@@ -1,63 +1,59 @@
...
-    (rm -rf var || true) \
-    # Create the var sub-directories
-    && mkdir -p var/cache var/logs var/sessions \
-    # Install dependencies
-    && composer install --prefer-dist --no-scripts --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative \
-    # Fixes permissions issues in non-dev mode
-    && chown -R www-data . var/cache var/logs var/sessions
-
-CMD ["/app/docker/start.sh"]
+	(rm -rf var || true) \
```
teohhanhui: This line does nothing useful and can be removed. :)

docker-compose.yml
teohhanhui: This is just giving the volume a name. :smile: The official MySQL image already uses a VOLUME instruction for the data directory.

But @jockos I'd match the name of the service, so "database-data". Since it was decided(?) not to rename it to the shorter / more convenient "db"...

dunglas: > Since it was decided(?) not to rename it to the shorter / more convenient "db"...

Symfony use `database_*` as parameters. It's better to stay consistent.

dunglas: But as the official image already use a volume, I would remove this entry entirely.

Dockerfile
Simperfit: Why two runs ?

.travis.yml
```diff
@@ -5,5 +5,6 @@ services:
 
 script:
     - docker-compose build
-    - docker-compose run --rm web composer install -o -n
-    - docker-compose run --rm web bin/console security:check
+    - docker-compose run --rm php composer install -o -n
```
Simperfit: the return of  `docker-compose exec` may be strange sometime with travis.

jockos:
@teohhanhui I have made some changes according to your review, can you check again ? :)

docker-compose.yml
teohhanhui: What Symfony uses does not affect / should not inform how we name Docker Compose services. :smile:

teohhanhui: https://docs.docker.com/engine/tutorials/dockervolumes/#/creating-and-mounting-a-data-volume-container

The Docker documentation recommends using data volume containers, but it's outdated. Named data volumes should be used instead. See https://github.com/docker/docker/issues/20465

I think we should keep this.

Dockerfile
```diff
@@ -1,63 +1,57 @@
...
-    && docker-php-ext-enable --ini-name 20-apcu.ini \
-        apcu
-
-# Apache config
-RUN a2enmod rewrite
-ADD docker/apache/vhost.conf /etc/apache2/sites-available/000-default.conf
+FROM php:7.1-fpm-alpine
+
+RUN apk add --no-cache --virtual .run-deps \
+		git \
```
teohhanhui: Oops... Double tabs.

Dockerfile
```diff
@@ -1,63 +1,57 @@
...
-ADD . /app
-WORKDIR /app
+COPY composer.json ./
+COPY composer.lock ./
 
-# Fix permissions (useful if the host is Windows)
-RUN chmod +x docker/composer.sh docker/start.sh docker/apache/start_safe_perms
+COPY app app/
+COPY bin bin/
+COPY docker/composer.sh /usr/local/bin/docker-app-install-composer
```
teohhanhui: Rename docker/composer.sh to docker/php/composer.sh to be consistent. :)

Dockerfile
```diff
@@ -1,63 +1,57 @@
...
+COPY src src/
+COPY web web/
+
+COPY docker/php/start.sh /usr/local/bin/docker-app-start
```
teohhanhui: Just combine this into the same visual block of instructions above, and sort alphabetically.

Dockerfile
```diff
@@ -1,63 +1,57 @@
...
+
+RUN chmod +x /usr/local/bin/docker-app-start /usr/local/bin/docker-app-install-composer
```
teohhanhui:
```Dockerfile
RUN chmod +x \
	/usr/local/bin/docker-app-install-composer \
	/usr/local/bin/docker-app-start
```

Dockerfile
```diff
@@ -1,63 +1,57 @@
...
-    # Create the var sub-directories
-    && mkdir -p var/cache var/logs var/sessions \
-    # Install dependencies
-    && composer install --prefer-dist --no-scripts --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative \
-    # Fixes permissions issues in non-dev mode
-    && chown -R www-data . var/cache var/logs var/sessions
-
-CMD ["/app/docker/start.sh"]
+RUN docker-app-install-composer \
+	&& mv composer.phar /usr/bin/composer \
```
teohhanhui: This should be /usr/local/bin/composer

https://getcomposer.org/doc/00-intro.md#globally

Dockerfile
```diff
@@ -1,63 +1,57 @@
...
+	&& su-exec www-data composer global require "hirak/prestissimo:^0.3" \
+	&& su-exec www-data composer clear-cache
+
+RUN chown -R www-data .
+
+RUN mkdir -p var/cache var/logs var/sessions \
```
teohhanhui:
```Dockerfile
RUN mkdir -p \
	var/cache \
	var/logs \
	var/sessions
```

docker-compose.yml
```diff
@@ -1,18 +1,31 @@
...
+            MYSQL_USER: api_platform
+        image: mysql:5.7
+    nginx:
+        image: nginx:1.11-alpine
+        depends_on:
+            - php
+        ports:
+            - "80:80"
+        volumes:
+            - ./:/srv/app
```
teohhanhui: There is no reason why nginx needs access to the entire project directory. I suggest we change in composer.json:

```json
{
    ...
    "extra": {
        ...
        "symfony-assets-install": "hard",
        ...
    }
    ...
}
```

and mount the web directory only.

docker-compose.yml
```diff
@@ -1,18 +1,31 @@
...
-
-database:
-    image: mysql
-    environment:
-        MYSQL_ROOT_PASSWORD: changeme
-        MYSQL_DATABASE: api_platform
-        MYSQL_USER: api_platform
-        MYSQL_PASSWORD: api_platform
+version: '2'
+services:
```
teohhanhui: Let's add back the named volume as mentioned in https://github.com/api-platform/api-platform/pull/162#discussion_r100718173

jockos:
Ping @teohhanhui :)

dunglas:
Travis is not happy @jockos

teohhanhui:
@dunglas That was my fault. I think it should be all fixed now.

Dockerfile
teohhanhui: We've got the best of both worlds now. :smile:

teohhanhui:
Travis failure (`bin/console security:check`) seems to be a problem on SensioLabs side.

teohhanhui:
@jockos Do verify that everything is working.

Lctrs:
To help solving permissions hell once and for all, Have you read this blog post : https://medium.com/@rawkode/avoiding-docker-compose-entry-points-with-volume-driver-options-f2b30c051ead ?

I used this for my dev environment and I never had a permission problem anymore (on Linux, haven't tested it on Mac or Windows).

For the record, here are my docker-compose.yml and my Dockerfile for a dev environment for a project that does NOT include ApiPlatform.

docker-compose.yml : 
```yaml
version: "2"

volumes:
    cache:
        driver: local
        driver_opts:
            type: tmpfs
            device: tmpfs
            o: uid=33
    logs:
        driver: local
        driver_opts:
            type: tmpfs
            device: tmpfs
            o: uid=33
    sessions:
        driver: local
        driver_opts:
            type: tmpfs
            device: tmpfs
            o: uid=33

services:
    web:
        build: ./docker/web
        environment:
            DNSDOCK_ALIAS: www.myproject.dev
            SYMFONY_ENV: dev
        depends_on:
            - db
        volumes:
            - .:/srv/app
            - cache:/srv/app/var/cache
            - logs:/srv/app/var/logs
            - sessions:/srv/app/var/sessions

    db:
        image: postgres:9.4
        environment:
            POSTGRES_PASSWORD: myproject
            POSTGRES_USER: myproject
            POSTGRES_DB: myproject
```

Dockerfile (located in docker/web) :
 ```Dockerfile
FROM php:7.1-apache

# PHP extensions
ENV APCU_VERSION 5.1.7

RUN buildDeps=" \
        libicu-dev \
        libpq-dev \
    " \
    && apt-get update \
    && apt-get install -y --no-install-recommends \
        $buildDeps \
        libicu52 \
        libpq5 \
    && rm -rf /var/lib/apt/lists/* \
    && docker-php-ext-install -j$(nproc) \
        intl \
        pdo_pgsql \
    && pecl install \
        apcu-$APCU_VERSION \
    && docker-php-ext-enable --ini-name 05-opcache.ini \
        opcache \
    && docker-php-ext-enable --ini-name 20-apcu.ini \
        apcu \
    && apt-get purge -y --auto-remove $buildDeps

# Apache & PHP configuration
RUN a2enmod rewrite
COPY apache/vhost.conf /etc/apache2/sites-enabled/000-default.conf
COPY php/php.ini /usr/local/etc/php/php.ini

WORKDIR /srv/app
```

(Also, there is no need to install mbstring extension since it's included in the parent image for a while now. See https://github.com/docker-library/php/blob/9abc1efe542b56aa93835e4987d5d4a88171b232/7.1/fpm/alpine/Dockerfile#L113)

**Comment by soyuka** [comment]
> Our fpm container is running as user www-data, which is uid 33

This isn't correct here, but assuming this sets the correct permissions to the volume inside the container, will you be able to write in the volume from the host? For the development version, we want to be able to edit files from the host.

**Comment by teohhanhui** [comment]
Using tmpfs means data is lost on container stop. I don't think that's a reasonable tradeoff for a development environment... (That is assuming it works not only on Linux. Because there is already a way that works perfectly for filesystems that supports acl - only Linux so far, until osxfs can support acl.)

**Comment by teohhanhui** [comment]
@Gormador @renan-taranto @Linkinou @Stoakes

I'd love to know if this fixes #192.

**Comment by Gormador** [comment]
I'm sorry I won't be able to give you feedback on this one as I moved away from API Platform for other reasons, and am therefor not in a position to test it.
Thank you for your efforts on this project, though! :-)

**Comment by dunglas** [comment]
@Gormador not related but can you share the other reasons? We're interested in this kind of feedback. Thanks.

**Comment by Gormador** [comment]
@dunglas Well, it appeared that using a JS framework to work with MongoDB simplified matters quite a bit... So I'm now using Express with Mongoose, which works fine (and made me learn JS, which is good!). However, I regret the native docker integration... And I'll be sure to check back on API Platform if I ever need to develop a php-based application!

**Comment by dunglas** [comment]
Thanks for the feedback. Indeed switching the "server" part to use Mongo or Firebase directly can be a good solution in some cases.

**Comment by jockos** [comment]
@teohhanhui LGTM, I've tested your modifications on Mac, Windows and Linux :)

File: docker/php/install-composer.sh [hunk 61]
```diff
@@ -1,16 +1,17 @@
 #!/bin/sh
-EXPECTED_SIGNATURE=$(php -r "echo trim(file_get_contents('https://composer.github.io/installer.sig'));")
+
+EXPECTED_SIGNATURE=$(wget -q -O - https://composer.github.io/installer.sig)
```
teohhanhui: I synced this with the latest snippet from https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

dunglas: @teohhanhui can you revert this? As we use PHP 7+, we can use PHP directly for the download and remove wget from the container: https://github.com/composer/composer/pull/5860

teohhanhui: We don't have wget in the container. It's the the wget from BusyBox. :smile:

I don't think this is something important enough worth diverging from the official install script for...

dunglas: Ok for me. Can we merge?

teohhanhui: Yes please.

dunglas:
Thank you very much @jockos @teohhanhui @soyuka!

backbone87:
using Docker for WIndows with the master branch works
(2.0.7 does NOT work in Docker for Windows due to complaints of apache to not start up as root)

also i created a modified version which uses postgres insteadof mysql,
i could create a PR for it, but i dont know if it makes sense to provide multiple docker container configs

PR #190, opened 2016-12-03
Upgrade HautelookAliceBundle

theofidry:
The new alice version is not released as stable yet but the user-land usage should be rather stable  and I really recommend to use the newer version as there is very little support for the old one. It will probably be released as stable in a couple of months so it makes sense to recommend the beta now.


dunglas:
Closing as it should no be installed with Flex. (will be part of the demo)

PR #203, opened 2017-01-05
Remove pre-existing PID file before starting apache

krismas:
In some cases when stopped, container running apache leaves a PID file behind. As a result, the next time the container starts, apache will stop because of the PID file left in place : 

```
httpd (pid 48) already running
```


dunglas:
@krismas thanks for your contribution, can you rebase against 2.0 please as this is a bug fix (I've already changed the target branch in the PR for you).

krismas:
my rebase failed !

PR #212, opened 2017-01-13
Use --classmap-authoritative for Composer install in prod mode

dunglas:


Dockerfile
```diff
@@ -53,7 +53,7 @@ RUN \
     # Create the var sub-directories
     && mkdir -p var/cache var/logs var/sessions \
     # Install dependencies
-    && composer install --optimize-autoloader --no-scripts \
+    && composer install --prefer-dist --no-scripts --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
```
Simperfit: missing final `\`

dunglas: fixed

PR #215, opened 2017-01-18
Improve Docker support on Windows

### LINKED ISSUE #213: Install on Windows 7 using docker toolbox
Dockerfile
Add in 
RUN chmod +x ./docker/composer.sh
under
 Install composer
and before 
RUN ./docker/composer.sh \

Need to figure out path permissions still

dunglas:
Closes #213.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,9 @@ RUN apt-get update \
 ADD . /app
 WORKDIR /app
 
+# Fix permissions (useful if the host is Windows)
+RUN chmod +x docker/composer.sh docker/start.sh docker/apache/start_safe_perms
+
 # Install composer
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -43,7 +43,8 @@ ADD . /app
 WORKDIR /app
 
 # Install composer
-RUN ./docker/composer.sh \
+RUN chmod +x ./docker/composer.sh \
```
teohhanhui: Don't we have to do the same for all other .sh files?

dunglas: Done

dunglas:
See also https://github.com/api-platform/docs/pull/172

Issue #192, opened 2016-12-15
Can't access localhost/app_dev.php (similar to #169) (docker install, Ubuntu 16.04)


Gormador:
Hi,

Currently banging my head against a wall of incomprehension at this stage of the [tutorial](https://api-platform.com/docs/distribution/):
![appdevfails](https://cloud.githubusercontent.com/assets/6118753/21233871/1c266fac-c2f1-11e6-9c22-54a58beadc95.png)

What I did:
- corrected my mistake of not having used `docker-compose` as a user instead of root by adding my user to the docker group, rebooted, deleted the already built docker images, deleted the skeleton-project including the root-owned folders;
- ran `docker-compose up -d` again, as a user this time, in a freshly copy pasted skeleton-project and followed the tutorial up to the same point;
- same error. `var/cache/dev` is still root-owned.

No idea what I did wrong. Any clue as to what it might be?
Thanks in advance for any input you may have, and thank you for this great looking project :-)

ps: May be worth noting that I'm a docker noob.

Simperfit:
the /var/cache is root locally too ?

Could you please try to remove the whole folders content give the permission to the current user ?

Gormador:
Yes, sorry, didn't mention it. I'm speaking about the `var/cache/dev` folder in the skeleton-project directory.

Trying out your suggestion now...

Ok, performed `sudo chown -R $USER var/cache/dev` after deleting its content. Refreshing the page gives me an Error 500. But applying the same to `var/cache/logs/dev.log` fixed it and allowed me to access the dev env.

I have no idea why this behaved like it did though. Why did root took ownership of those files when the command I was using wasn't executed as root?

Anyway, I can at least continue with the tutorial now.

Thanks!

Simperfit:
I think something has been executing as root at the first time

renan-taranto:
I'm also having this problem.

- I didn't run docker-compose as root

renan-taranto:
I'm also having this problem.

- I didn't run `docker-compose` as root;
- I had to `sudo chown -R $USER var/logs/` and `sudo chown -R $USER var/cache/` to access the dev env.

IMHO this issue should not be closed since it wasn't fixed.

teohhanhui:
The user in the container which executed the commands is `root`, so that might be the problem... (I haven't looked into it.)

Linkinou:
By default, Docker containers are executed as root.
So if you execute a command through a docker container, it will be executed by the root User.

[...]

PR #218, opened 2017-01-23
Solve cache not writable on docker

Stoakes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #192
| License       | MIT
| Doc PR        |  x

On docker run, start.sh launch a composer install as root user. Therefore, apache user can't read var/cache/dev content.

To fix it, I get apache user and apache group in `/etc/apache2/apache2.conf` and chown var folder with those values.

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -10,6 +10,13 @@ else
     composer install --prefer-dist --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
 fi
 
+#get apache user : second word of the first line (from docker/apache/start_safe_perms)
+apache_user=$(head -n 1 /etc/apache2/apache2.conf| cut -d " " -f 2)
+#get apache group : second word of the second line
+apache_group=$(head -n 2 /etc/apache2/apache2.conf| tail -1 |  cut -d " " -f 2)
+
+chown -R $apache_user:$apache_group var
+
 # Start Apache with the right permissions after removing pre-existing PID file
 rm -f /var/run/apache2/apache2.pid
 exec docker/apache/start_safe_perms -DFOREGROUND


soyuka:
Not sure that this should be fixed from the `start.sh` file, see [this file](https://github.com/Stoakes/api-platform/blob/163969f52bb29fd72e0724c6666473ec0978694c/docker/apache/start_safe_perms)

docker/start.sh
```diff
@@ -7,9 +7,16 @@ export DOCKER_BRIDGE_IP=$(ip ro | grep default | cut -d' ' -f 3)
 if [ "$SYMFONY_ENV" = 'dev' ]; then
     composer install --prefer-dist --no-progress --no-suggest
 else
-    composer install --prefer-dist --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
+    composer install --prefer-dist --no-dev --no-progress --no-suggest --optimize-autoloader
```
Simperfit: why did you remove `--classmap-authoritative` ?

docker/start.sh
```diff
@@ -7,9 +7,16 @@ export DOCKER_BRIDGE_IP=$(ip ro | grep default | cut -d' ' -f 3)
...
 fi
 
+#get apache user : second word of the first line (from docker/apache/start_safe_perms)
```
Simperfit: It does not fix the issue when we are using `docker-compose exec web composer update`

docker/start.sh
Stoakes: My mistake,
Copy/paste error.

theofidry: Indeed it shouldn't be change, `--optimize-autoloader` can be removed though as `--classmap authoritative` does imply `--optimize-autoloader`.

Stoakes: Solved.

docker/start.sh
Stoakes: Sure.

But except with --user flag, I think there is no way to run a command as a non-root user.

dunglas:
Wouldn't be better to run the composer install with the apache user instead of doing this?

docker/start.sh
```diff
@@ -10,6 +10,13 @@ else
     composer install --prefer-dist --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
 fi
 
+#get apache user : second word of the first line (from docker/apache/start_safe_perms)
+apache_user=$(head -n 1 /etc/apache2/apache2.conf| cut -d " " -f 2)
```
dunglas: It cannot work because the apache user and group is changed after that: https://github.com/Stoakes/api-platform/blob/163969f52bb29fd72e0724c6666473ec0978694c/docker/apache/start_safe_perms#L26-L32

soyuka:
> Wouldn't be better to run the composer install with the apache user instead of doing this?
 
+1 everything should be done as one user, ie `www-data`.

Stoakes:
Running everything as www-data would be easier.

In fact I don't know why api-platform does all those stuff around owner and group, I was just trying to fix the issue.
 As @soyuka pointed out, it would be easier (and less hacky) to manage permissions in start_safe_perms.

So what's next ? Do I update this PR or close it in favour of a whole docker system rewrite ?

dunglas:
@Stoakes because we use volumes for dev, and we cannot change the owner of files mounted as a volume.

soyuka:
Hmm, I don't get why this is needed.

What I do in my own config:

```
version: '2'
services:
  php:
    container_name: api_platform_php
    build: php-fpm
    ports:
      - "9000:9000"
    volumes:
      - ./symfony:/var/www/symfony
  nginx:
    container_name: api_platform_nginx
    build: nginx
    ports:
      - "8081:80"
    depends_on:
      - php
    volumes:
      - ./symfony:/var/www/symfony
      - ./logs/nginx/:/var/log/nginx
```

Where `./symfony` is a symlink to the project:

```
lrwxrwxrwx 1 abluchet abluchet   21 Aug  1 17:40 symfony -> /home/abluchet/project
```

Then, everything runs as `www-data`, I have no hacks whatsoever. The only thing I do in both Dockerfiles is to set the same `uid`:

```
RUN usermod -u 1000 www-data
```

I could post my docker configuration if you want.

dunglas:
@soyuka are you using Linux or Mac?

soyuka:
Both linux @ work and os x @ home. This works on both.

dunglas:
Yes but with our trick we doesn't need to set the uid.

soyuka:
You mean using a 50 lines bash script instead of 1 line in the Dockerfile? ^^ Anyway it's different here because I just need nginx and php-fpm to run on the same user. Though I don't need anything for my volume permission.

dunglas:
@soyuka it's very important for the first impression to have something working out of the box, without requiring any weird config, even for people not confortable with Docker.

soyuka:
> @soyuka it's very important for the first impression to have something working out of the box, without requiring any weird config, even for people not confortable with Docker.

100% agreed

soyuka:
Okay, so I've spent a few hours playing with your docker configuration. I see a few issues but those can not be solved properly with the current  `FROM php:7.0-apache`.

The hack in `start_safe_perm` is not necessary (oh and btw it comes from [here](https://github.com/kojiromike/docker-magento/blob/master/apache/start_safe_perms)) if we use the `envvars` file that is sourced by `apache2-foreground`, a binary declared by https://github.com/docker-library/php. 

We could make this work if we were not installing the composer dependencies within docker. Indeed, by doing so, if we want correct permissions, we need to issue the `composer` command from `www-data`. If it's run by `root` directories will have the wrong permissions. For example, as `composer` post install scripts are creating the cache directories, it'll fail.

So what about running everything as `www-data` as I suggested?

 Because, the docker source image we're using is exposing the `80` port, which is restricted to root. Though, it would just work great with the `8080` port which would be forwarded to `80` by docker-compose (host has root capabilities). The thing is, it's hard coded and there are no ways to change it (check the related issues on https://github.com/docker-library/php). 

What now?

A cleaner hack is to use a simple `su -s /bin/bash -c "composer command" www-data` when we need to so that the permissions are maintained.

IMHO, it would be best to make this work by using `php-fpm` and `nginx`, because they're not using any unsafe port. Or, we just use another `php` image that doesn't require the `80` port ([example with alpine + nginx](https://github.com/xataz/dockerfiles/blob/master/nginx-php/Dockerfile)).

I'm adding a pull-request now.

dunglas:
@soyuka as we're in the process of merging #162 who's already using FPM, can you push commits (or comment) in this branch instead?

soyuka:
Okay I'll take a look at it!

Simperfit:
Since #162 has been merge, it is still relevant ?

PR #221, opened 2017-01-25
Better docker permissions managment

soyuka:
https://github.com/api-platform/api-platform/pull/218#issuecomment-275159435

Can someone try this on windows? Thanks!


soyuka:
Wow, I really don't get where does the `vendor` directory goes. Everything goes well :

```
Writing /var/www/.composer/cache/files/symfony/swiftmailer-bundle/eed9cefd9dccaf9e2407c55d0f15ad1507da4754.zip into cache from /var/www/app/vendor/symfony/swiftmailer-bundle/b1aeaaefeb5385689d093a4766210939
```

But the vendor directory just doesn't get written. Though, when doing `docker-compose up -d` you can access `localhost` and everything is fine.

PR #222, opened 2017-01-26
Always use the dev env if the SYMFONY_ENV var isn't set

dunglas:
Otherwise the `composer install` fail because `dev` is the env by default of Symfony but dev dependencies aren't installed.

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -4,10 +4,10 @@ set -xe
 # Detect the host IP
 export DOCKER_BRIDGE_IP=$(ip ro | grep default | cut -d' ' -f 3)
 
-if [ "$SYMFONY_ENV" = 'dev' ]; then
-    composer install --prefer-dist --no-progress --no-suggest
-else
+if [ "$SYMFONY_ENV" = 'prod' ]; then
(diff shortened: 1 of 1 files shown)


docker/start.sh
```diff
@@ -4,10 +4,10 @@ set -xe
 # Detect the host IP
 export DOCKER_BRIDGE_IP=$(ip ro | grep default | cut -d' ' -f 3)
 
-if [ "$SYMFONY_ENV" = 'dev' ]; then
-    composer install --prefer-dist --no-progress --no-suggest
-else
+if [ "$SYMFONY_ENV" = 'prod' ]; then
```
teohhanhui: I think "not dev" would be better.

dunglas: Not really, in test env we should install dev dependencies too. IMO we should not install dev deps only in prod mode.

teohhanhui:
The same could be said about "staging" or other environments...

dunglas:
But there is no staging env by default (only test, dev and prod AFAIK). If an user add such env, it's its responsibility to update the script (and usually a bad idea to have a "staging" env, it should just use "prod").

PR #223, opened 2017-01-26
Update config_prod.yml settings

dunglas:
* Remove the comment regarding the framework section (not necessary anymore since 3.2)
* Switch to the apcu driver of Doctrine Cache (best default for PHP 7+ because it doesn't need the compatibility layer of APCu).

### DIFF
--- a/app/config/config_prod.yml
+++ b/app/config/config_prod.yml
@@ -1,17 +1,11 @@
 imports:
     - { resource: config.yml }
 
-#framework:
-#    validation:
-#        cache: validator.mapping.cache.doctrine.apc
-#    serializer:
-#        cache: serializer.mapping.cache.doctrine.apc
-
-#doctrine:
-#    orm:
-#        metadata_cache_driver: apc
-#        result_cache_driver: apc
-#        query_cache_driver: apc
+doctrine:
+    orm:
+        metadata_cache_driver: apcu
+        result_cache_driver: apcu
+        query_cache_driver: apcu
 
 monolog:
     handlers:


app/config/config_prod.yml
```diff
@@ -1,17 +1,11 @@
...
-#        cache: validator.mapping.cache.doctrine.apc
-#    serializer:
-#        cache: serializer.mapping.cache.doctrine.apc
-
 #doctrine:
 #    orm:
-#        metadata_cache_driver: apc
-#        result_cache_driver: apc
-#        query_cache_driver: apc
+#        metadata_cache_driver: apcu
```
dunglas: Maybe can we enable such optim by default because our Docker setup comes with APCu. What do you think?

soyuka: Can we enabled those in the production environment only? Can be tricky if you don't know that apcu cache is running and you're wondering why some things don't update.

dunglas: `config_prod.yml` is only parsed in the production env.

soyuka: oh... Then yes ofc.

app/config/config_prod.yml
```diff
@@ -1,17 +1,11 @@
...
-#        cache: validator.mapping.cache.doctrine.apc
-#    serializer:
-#        cache: serializer.mapping.cache.doctrine.apc
-
-#doctrine:
-#    orm:
-#        metadata_cache_driver: apc
-#        result_cache_driver: apc
-#        query_cache_driver: apc
+doctrine:
```
teohhanhui: Shouldn't this section remain commented out? The user needs to make a conscious decision here.

dunglas: The driver should always be configured like this if APCu is enabled. It will improve performances thanks to the metadata cache but doesn't have any side effect because even with a configured driver, result and query cache aren't enabled by default.

The only downside of uncommenting them is when APCu isn't installed. But it is in our Docker setup (the recommended way to install API Platform).

teohhanhui: Actually, I thought our policy is to stick as close to Symfony Standard as possible?

If so, we should follow: https://github.com/symfony/symfony-standard/blob/v3.2.2/composer.json#L21 (symfony/symfony-standard#974)

Probably this file should be changed in Symfony Standard as well.

PR #224, opened 2017-01-27
Docker nginx + php-fpm on alpine

soyuka:
:man_dancing: 

If you want to test this:

```
docker-compose up -d
docker-compose run --rm php chown -R $(id -u):$(id -g) . # permission fix on the volume
```

Btw, I think that my previous implementation based on `alpine` was slightly faster because it didn't require to compile `intl, mbstring, pdo_mysql, zip` but rather installed those packages from alpine registry. 
This however is based on `php:7.0-fpm-alpine`.

I added some lines in the travis configuration to test permissions, let's see how it'll handle those.


Dockerfile
```diff
@@ -1,63 +1,34 @@
...
-        libicu-dev \
-        zlib1g-dev \
-    " \
-    && apt-get update \
-    && apt-get install -y --no-install-recommends \
-        $buildDeps \
-        libicu52 \
-        zlib1g \
-    && rm -rf /var/lib/apt/lists/* \
+FROM php:7.0-fpm-alpine
```
dunglas: `FROM php:7.1-fpm-alpine`?

Dockerfile
```diff
@@ -1,63 +1,34 @@
...
+
+# Git is for composer
+# openssh for future security support see https://github.com/api-platform/api-platform/issues/109
```
dunglas: We need openssl, not openssh. IMO We should keep the image minimal and let the user install extra packages only if he need them.

Dockerfile
```diff
@@ -1,63 +1,34 @@
...
+        ${BUILD_DEPS} \
+        git openssh su-exec \
+        icu-libs zlib \
     && docker-php-ext-install \
-        intl \
-        mbstring \
-        pdo_mysql \
-        zip \
-    && apt-get purge -y --auto-remove $buildDeps
-RUN pecl install \
```
dunglas: We still need APCu and the opcache extension.

Dockerfile
```diff
@@ -1,63 +1,34 @@
...
+# speed up composer
+RUN su-exec www-data composer global require hirak/prestissimo:^0.3
 
-RUN \
-    # Remove var directory if it's accidentally included
-    (rm -rf var || true) \
-    # Create the var sub-directories
-    && mkdir -p var/cache var/logs var/sessions \
-    # Install dependencies
-    && composer install --prefer-dist --no-scripts --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative \
```
dunglas: I prefer keeping the composer install in the image. It speeds up deployments because packages are already in the Composer cacher.

Dockerfile
soyuka: We may, wasn't sure I'll change this.

Dockerfile
soyuka: Okay, should I remove it then?

Dockerfile
dunglas: Yes please

Dockerfile
dunglas: Yes please

soyuka:
apcu and opcache added!
Composer installation is done on the build phase. It forces us to copy everything while we left the volume do its thing without it.

Dockerfile
```diff
@@ -1,63 +1,44 @@
...
+# speed up composer
+RUN su-exec www-data composer global require hirak/prestissimo:^0.3
+
+WORKDIR /api-platform
+
+COPY . /api-platform/
+
+RUN chown -R www-data .
+
+RUN su-exec www-data composer install --prefer-dist --no-scripts --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
```
teohhanhui: I do not recommend running composer install after copying all project files. It completely destroys caching whenever any file is changed.

Dockerfile
```diff
@@ -1,63 +1,44 @@
...
+
+# prepare volume directory
+RUN mkdir /api-platform
+
+# speed up composer
+RUN su-exec www-data composer global require hirak/prestissimo:^0.3
+
+WORKDIR /api-platform
+
+COPY . /api-platform/
```
teohhanhui: I'm still against copying everything blindly.

Dockerfile
```diff
@@ -1,63 +1,44 @@
...
-        libicu52 \
-        zlib1g \
-    && rm -rf /var/lib/apt/lists/* \
+
+# Git is for composer
+# su-exec for managing permissions
+RUN export BUILD_DEPS="zlib-dev icu-dev autoconf build-base" \
+    && apk --update --no-cache add \
+        ${BUILD_DEPS} \
+        git su-exec \
```
teohhanhui: One per line please, and sorted alphabetically.

Dockerfile
```diff
@@ -1,63 +1,44 @@
...
-    && apt-get install -y --no-install-recommends \
-        $buildDeps \
-        libicu52 \
-        zlib1g \
-    && rm -rf /var/lib/apt/lists/* \
+
+# Git is for composer
+# su-exec for managing permissions
+RUN export BUILD_DEPS="zlib-dev icu-dev autoconf build-base" \
+    && apk --update --no-cache add \
```
teohhanhui: The standard pattern is:

```Dockerfile
	apk add --no-cache --virtual .build-deps \
```

Also, please do a separate `apk add` for `.run-deps`.

teohhanhui: Use tabs for Dockerfile and shell scripts, following CS of Docker official images.

Dockerfile
teohhanhui: This is what I'd suggest:

```Dockerfile
WORKDIR /srv/app

COPY composer.json ./
COPY composer.lock ./

RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest \
	&& composer clear-cache

COPY app app/
COPY bin bin/
COPY src src/
COPY web web/
```

docker-compose.yml
```diff
@@ -1,18 +1,29 @@
...
-        SYMFONY_ENV: dev
-    links:
-        - database
-    volumes:
-        - .:/app
-    ports:
-        - 80:80
+version: '2'
+services:
+    php:
```
teohhanhui: Suggest alphabetical sorting.

Dockerfile
```diff
@@ -1,63 +1,44 @@
...
+
+COPY . /api-platform/
+
+RUN chown -R www-data .
+
+RUN su-exec www-data composer install --prefer-dist --no-scripts --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
+
+COPY ./docker/php/docker-entrypoint.sh /usr/local/bin/
+RUN chmod +x /usr/local/bin/docker-entrypoint.sh
+CMD ["docker-entrypoint.sh"]
```
teohhanhui: This is wrong. An ENTRYPOINT script is different from the start / CMD script.

Dockerfile
teohhanhui: @dunglas We do not want cache to be included in the layers of a Docker image (it becomes bloated). So the practice is to always clear cache after use (in the same RUN).

Dockerfile
soyuka: docker-cmd it'll be called then...

Dockerfile
soyuka: Why? Anyway I'm not for copying at all we should leave the volumes do their thing. This means that we don't `composer install`

Dockerfile
soyuka: Okay ` --no-autoloader --no-scripts` was obviously missing from my attempt to run it without the sources. What about I don't copy anything but the composer.json and let the volumes add the sources?

Dockerfile
soyuka: I didn't knew about virtual packages and will add them. Thanks!

Dockerfile
soyuka: I don't know who made that rule and I surely don't understand it (except for the alphabetical part).

Dockerfile
teohhanhui: I think the golden rule is for the image to be ready to run on container start. We need to strike the right balance, of course.

Dockerfile
teohhanhui: One package per line minimizes git diffs (helps to better track changes across commits)... And it's also easier to read. Well, the point is there's no downside to it.

Dockerfile
```diff
@@ -1,63 +1,48 @@
...
+RUN su-exec www-data composer global require hirak/prestissimo:^0.3
+
+WORKDIR /api-platform
+
+COPY composer.json .
+COPY composer.lock .
+
+RUN chown -R www-data .
+
+RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --no-suggest \
```
teohhanhui: I think this is missing `su-exec www-data ...`?

Dockerfile
soyuka: If this is the case, there's no need to install composer in the build phase no?

Dockerfile
```diff
@@ -1,63 +1,58 @@
...
+	&& mv composer.phar /usr/bin/composer \
+	&& chmod +x /usr/bin/composer
+
+# prepare volume directory
+RUN mkdir /api-platform
+
+# speed up composer
+RUN su-exec www-data composer global require hirak/prestissimo:^0.3
+
+WORKDIR /api-platform
```
teohhanhui: I suggest we use `/srv/app`.

Dockerfile
teohhanhui: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/sort-multi-line-arguments

Dockerfile
teohhanhui: See my reply here: https://github.com/api-platform/api-platform/pull/224#discussion_r98375924

teohhanhui:
I'd like to see this and #162 combined.

Dockerfile
soyuka: I disagree, why `/srv`? Also why do we need a sub path if all we need is `/somedirectory`?

Dockerfile
```diff
@@ -1,63 +1,58 @@
...
+
+COPY composer.json .
+COPY composer.lock .
```
soyuka: @teohhanhui wdyt about only using the composer files instead of the whole library?

soyuka:
Differences with #162:

- api-platform root path `/src/app` instead of `/api-platform` (I really don't get this part)
- implement named volumes instead of volume_from
- COPY every sub-path of the root repository (what about using only the composer files and letting the volume do the rest?)

Dockerfile
teohhanhui: http://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#srvDataForServicesProvidedBySystem

jockos:
@soyuka look [here](http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/srv.html) for the `/srv` part.

teohhanhui:
> what about using only the composer files and letting the volume do the rest?

Ideally, the image should be self-sufficient. Of course, there are exceptions like the data directory of a database, where it's not really practical to do so. But I'd expect the image of an api-platform *app* to include the files that actually make up the *app*.

soyuka:
#162

PR #237, opened 2017-02-20
[Docker] add standard streams

20uf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

Hello,

I added in the docker file the output of logs in console, I find absent and yet very practical to debug.

Regards

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,4 +66,7 @@ COPY docker/php/start.sh /usr/local/bin/docker-app-start
 
 RUN chmod +x /usr/local/bin/docker-app-start
 
+RUN ln -sf /dev/stdout /var/log/nginx/app_access.log
+RUN ln -sf /dev/stderr /var/log/nginx/app_error.log
+
 CMD ["docker-app-start"]


dunglas:
Are you sure about your patch? Nginx logs are in the NGINX container, not in the PHP one. However maybe can we send Symfony logs to stderr.

20uf:
Hmm true.

Should logs be sent with monolog?

teohhanhui:
http://stackoverflow.com/a/38501354/1529493

But I think we should log to both `var/logs/*.log` and `stderr`.

20uf:
I don't think it's appropriate for monolog to do that.
This is as part of a use with Docker, so it would actually be at the nginx container to do so.

teohhanhui:
It needs to be done for each container. :smile:

For the Symfony logs, we can potentially do:

```sh
...
	&& ln -sf /dev/stderr var/logs/dev.log \
	&& ln -sf /dev/stderr var/logs/prod.log
```

See https://github.com/nginxinc/docker-nginx/blob/7b33a90d7441909664a920b0687db8d984ac314b/mainline/alpine/Dockerfile#L126-L128

PR #238, opened 2017-02-20
Add a Varnish container and enable cache invalidation

dunglas:
See api-platform/core#952 for the rationale.

TODO:
* [x] Disable the cache for `app_dev.php`

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -79,3 +79,13 @@ nelmio_cors:
         max_age: 3600
     paths:
         '^/': ~
+
+api_platform:
+    http_cache:
+        invalidation:
+            enabled: true
+            varnish_urls: '%varnish_urls%'
+        max_age: 0
+        shared_max_age: 3600
+        vary: ['Content-Type', 'Authorization']
+        public: true
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "api-platform/core": "2.1.x-dev",
         "doctrine/doctrine-bundle": "^1.6",
         "doctrine/orm": "^2.5",
+        "guzzlehttp/guzzle": "^6.0",
         "incenteev/composer-parameter-handler": "^2.0",
         "nelmio/cors-bundle": "^1.4",
         "phpdocumentor/reflection-docblock": "^3.0",
--- a/docker/varnish/conf/default.vcl
+++ b/docker/varnish/conf/default.vcl
@@ -1,0 +1,86 @@
+vcl 4.0;
+
+import std;
+
+backend default {
+  .host = "nginx";
+  .port = "80";
+  # Health check
+  .probe = {
+    .url = "/";
+    .timeout = 5s;
+    .interval = 10s;
+    .window = 5;
+    .threshold = 3;
+  }
+}
+
+# Hosts allowed to send BAN requests
+acl ban {
+  "localhost";
+  "app";
+}
+
+sub vcl_backend_response {
+  # Ban lurker friendly header
+  set beresp.http.url = bereq.url;
+
+  # Add a grace in case the backend is down
+  set beresp.grace = 1h;
+}
+
+sub vcl_deliver {
+  # Don't send cache tags related headers to the client
+  unset resp.http.url;
+  # Uncomment the following line to NOT send the "Cache-Tags" header to the client (prevent using CloudFlare cache tags)
+  #unset resp.http.Cache-Tags;
+}
+
+sub vcl_recv {
+  # Remove the "Forwarded" HTTP header if exists (security)
+  unset req.http.forwarded;
+
+  # To allow API Platform to ban by cache tags
+  if (req.method == "BAN") {
+    if (client.ip !~ ban) {
+      return(synth(405, "Not allowed"));
+    }
+
+    if (req.http.ApiPlatform-Ban-Regex) {
+      ban("obj.http.Cache-Tags ~ " + req.http.ApiPlatform-Ban-Regex);
+
+      return(synth(200, "Ban added"));
+    }
+
+    return(synth(400, "ApiPlatform-Ban-Regex HTTP header must be set."));
+  }
+
+  # Don't cache in dev mode
+  if (req.url ~ "^/app_dev.php") {
+    return(pass);
+  }
+}
+
+# From https://github.com/varnish/Varnish-Book/blob/master/vcl/grace.vcl
+sub vcl_hit {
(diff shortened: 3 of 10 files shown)


composer.json
```diff
@@ -18,13 +18,20 @@
...
+        {
+            "type": "vcs",
+            "url": "https://github.com/dunglas/core"
+        }
+    ],
     "require": {
         "php": ">=7.0",
         "symfony/symfony": "3.2.*",
-        "api-platform/core": "^2.0",
+        "api-platform/core": "dev-cache_tags",
```
Simperfit: Don't forget to change it when merging both

docker/varnish/conf/default.vcl
```diff
@@ -0,0 +1,86 @@
+vcl 4.0;
+
+import std;
+
+backend default {
+  .host = "nginx";
```
jderusse: beware, varnish resolves the IP of the domain at boot time and keep it till the end of process.
If the IP of nginx changes (for instance when the container restarts) varnish won't be able to route traffic to the new container

docker/varnish/conf/default.vcl
```diff
@@ -0,0 +1,86 @@
...
+    if (req.http.X-Ban-Regex) {
+      ban("obj.http.Cache-Tags ~ " + req.http.X-Ban-Regex);
+
+      return(synth(200, "Ban added"));
+    }
+
+    return(synth(400, "X-Ban-Regex HTTP header must be set."));
+  }
+
+  # Don't cache in dev mode
```
jderusse: Why? Shouldn't be an issue if the invalidator works well. 
If it don't, you'll not going on prod.

docker/varnish/conf/default.vcl
```diff
@@ -0,0 +1,86 @@
...
+}
+
+sub vcl_recv {
+  # Remove the "Forwarded" HTTP header if exists (security)
+  unset req.http.forwarded;
+
+  # To allow API Platform to ban by cache tags
+  if (req.method == "BAN") {
+    if (client.ip !~ ban) {
+      return(synth(403, "Not allowed"));
```
jderusse: What's about 405?

docker/varnish/conf/default.vcl
```diff
@@ -0,0 +1,86 @@
...
+
+  # Add a grace in case the backend is down
+  set beresp.grace = 1h;
+}
+
+sub vcl_deliver {
+  # Don't send cache tags related headers to the client
+  unset resp.http.url;
+  # Uncomment the following line to NOT send the "Cache-Tags" header to the client (prevent using CloudFlare cache tags)
+  #unset resp.http.Cache-Tags;
```
jderusse: I wonder if we shouldn't remove the `s-max-age` header too.
With this new tag feature, the app sets an arbitrary value because we known we'll be able to invalidate the varnish. Which is fine.

BUT what if an other unknown proxy is on the path? (see "proxy cache" in this old version of the documentation http://symfony.com/doc/2.2/book/http_cache.html#types-of-caches)

teohhanhui: We should use a custom header like `X-Reverse-Proxy-TTL` instead of abusing `s-max-age`.

docker/varnish/conf/default.vcl
dunglas: Do you know a workaround for this?

docker/varnish/conf/default.vcl
dunglas: It hurts the DX. When you change the code in dec, you want the new code to be executed even if the underlying resource has not been modified.

docker/varnish/conf/default.vcl
jderusse: Several people use a local nginx configured in proxy pass.
Varnish point to nginx 127.0.0.1, and nginx performs the DNS resolution on the fly.
Some other use a script in charge of reloading varnish when the I'm change.

None is ideal to me.

In dev environment, I just restart my container, in production we use a local nginx

app/config/config.yml
```diff
@@ -79,3 +79,9 @@ nelmio_cors:
         max_age:        3600
     paths:
         '^/': ~
+
+api_platform:
+    http_cache:
+        enable_tags: true
+        varnish_urls: %varnish_urls%
```
Shine-neko: Quote ?

docker/varnish/conf/default.vcl
dunglas: I suggest to let it as is for now. WDYT?

docker/varnish/conf/default.vcl
dunglas: For instance, when using CloudFlare, the `s-max-age` header should be set too. I propose to keep it as is for now too. We'll see when we'll get some feedback.

teohhanhui: We must not set an incorrect `s-max-age`. It's 100% dangerous. If the user needs to do that for CloudFlare, they should set it themselves (or we should provide some easy way of setting it).

docker/varnish/conf/default.vcl
teohhanhui: I use https://github.com/nigoroll/libvmod-dynamic.

dunglas:
Requires api-platform/core#1156

docker/varnish/conf/default.vcl
dunglas: I don't get your point, we don't set this value automatically, it's up to the user. We just provide a good default when using the Varnish image.

teohhanhui: Alright. I see that the default configuration doesn't set `maxage` and `s-maxage`, so we're safe. :smile:

https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L123

teohhanhui: So my suggestion is that we should use a custom header, e.g. `Reverse-Proxy-TTL`.

dunglas: My concern is that this custom header will not be taken into account by Cloudflare. Maybe can we set the smaxage as we currently do but remove it with Varnish before the request go outside our network?

teohhanhui: If CloudFlare uses `s-maxage`, let the user change the `shared_max_age` option that we already provide in the bundle config. It's not a safe default, and the default setup in api-platform/api-platform uses Varnish anyway, not CloudFlare.

PR #244, opened 2017-03-02
Modernising docker-compose configuration

rawkode:
Signed-off-by: David McKay <david@rawkode.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,43 +2,43 @@ version: '2.1'
 
 volumes:
   db_data:
-    driver: "local"
+    driver: 'local'
 
 services:
   db:
-    image: "mysql:5.7"
+    image: 'mysql:5.7'
     environment:
-      MYSQL_DATABASE: "api_platform"
-      MYSQL_USER: "api_platform"
-      MYSQL_PASSWORD: "api_platform"
-      MYSQL_RANDOM_ROOT_PASSWORD: "true"
+      MYSQL_DATABASE: 'api_platform'
+      MYSQL_USER: 'api_platform'
+      MYSQL_PASSWORD: 'api_platform'
+      MYSQL_RANDOM_ROOT_PASSWORD: 'true'
     volumes:
-      - "db_data:/var/lib/mysql"
+      - 'db_data:/var/lib/mysql'
     healthcheck:
-      test: ["CMD-SHELL", "mysql --user=api_platform --password=api_platform -e 'SHOW TABLES' api_platform"]
-      interval: "5s"
-      timeout: "1s"
+      test: ['CMD-SHELL', 'mysql --user=api_platform --password=api_platform -e "SHOW TABLES" api_platform']
+      interval: '5s'
+      timeout: '1s'
       retries: 10
 
   php:
-    build: "."
+    build: '.'
     depends_on:
       db:
-        condition: "service_healthy"
+        condition: 'service_healthy'
     environment:
-      SYMFONY_ENV: "dev"
+      SYMFONY_ENV: 'dev'
     volumes:
-      - "./:/srv/api-platform"
+      - './:/srv/api-platform'
 
   nginx:
-    image: "nginx:1.11-alpine"
+    image: 'nginx:1.11-alpine'
     depends_on:
       php:
-        condition: "service_started"
+        condition: 'service_started'
     ports:
-      - "80:80"
+      - '80:80'
     volumes_from:
       - php
     volumes:
-      - "./docker/nginx/conf.d:/etc/nginx/conf.d:ro"
+      - './docker/nginx/conf.d:/etc/nginx/conf.d:ro'
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,38 +1,44 @@
-version: '2'
+version: '2.1'
+
+volumes:
+  db_data:
+    driver: "local"
 
 services:
   db:
+    image: "mysql:5.7"
     environment:
-      MYSQL_DATABASE: api_platform
-      MYSQL_PASSWORD: api_platform
-      MYSQL_ROOT_PASSWORD: changeme
-      MYSQL_USER: api_platform
-    image: mysql:5.7
+      MYSQL_DATABASE: "api_platform"
+      MYSQL_USER: "api_platform"
+      MYSQL_PASSWORD: "api_platform"
+      MYSQL_RANDOM_ROOT_PASSWORD: "true"
     volumes:
-      - db-data:/var/lib/mysql
-  nginx:
-    image: nginx:1.11-alpine
+      - "db_data:/var/lib/mysql"
+    healthcheck:
+      test: ["CMD-SHELL", "mysql --user=api_platform --password=api_platform -e 'SHOW TABLES' api_platform"]
+      interval: "5s"
+      timeout: "1s"
+      retries: 10
+
+  php:
+    build: "."
     depends_on:
-      - php
+      db:
+        condition: "service_healthy"
+    environment:
+      SYMFONY_ENV: "dev"
+    volumes:
+      - "./:/srv/api-platform"
+
+  nginx:
+    image: "nginx:1.11-alpine"
+    depends_on:
+      php:
+        condition: "service_started"
     ports:
       - "80:80"
+    volumes_from:
+      - php
     volumes:
-      - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
-      - ./web:/srv/api-platform/web
-  php:
-    build: .
-    depends_on:
-      - db
-    environment:
-      SYMFONY_ENV: dev
-    volumes:
-      - ./:/srv/api-platform
-      - ./var/cache:/srv/api-platform/var/cache
-      - ./var/logs:/srv/api-platform/var/logs
-      - ./var/sessions:/srv/api-platform/var/sessions
-      - ./var:/srv/api-platform/var
-      - ./vendor:/srv/api-platform/vendor
-      - ./web:/srv/api-platform/web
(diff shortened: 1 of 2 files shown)


rawkode:
I would have preferred that the `HEALTHCHECK` used the variables, `${MYSQL_USER}`, etc; but I believe this is a bug in the `docker daemon`.

We can update this if/when they approve my PR: https://github.com/docker/docker/pull/31491

rawkode:
Looks like Travis only supports up to docker-compose 1.8 :frowning:

teohhanhui:
You can download docker-compose yourself. :) I think that's mentioned in
the Travis documentation.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
-version: '2'
+version: '2.1'
+
+volumes:
+  db_data:
+    driver: "local"
```
dunglas: Can you use single quotes everywhere please?

rawkode:
Updated / Resolved, @dunglas @teohhanhui 

Thanks :+1:

dunglas:
Thanks @rawkode

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
+      SYMFONY_ENV: 'dev'
     volumes:
-      - ./:/srv/api-platform
-      - ./var/cache:/srv/api-platform/var/cache
-      - ./var/logs:/srv/api-platform/var/logs
-      - ./var/sessions:/srv/api-platform/var/sessions
-      - ./var:/srv/api-platform/var
-      - ./vendor:/srv/api-platform/vendor
-      - ./web:/srv/api-platform/web
+      - './:/srv/api-platform'
```
Simperfit: why removing the volumes ?

rawkode: They were replaced by a single mount, rather than a bunch within the same directory

teohhanhui:
Did not approve this... I'd have asked for lots of changes. :disappointed:

dunglas:
Sorry I missed your review. We can still revert.

dunglas:
But your review doesn't appear.

rawkode:
What changes would you have asked for, @teohhanhui ?

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
+      MYSQL_DATABASE: 'api_platform'
+      MYSQL_USER: 'api_platform'
+      MYSQL_PASSWORD: 'api_platform'
+      MYSQL_RANDOM_ROOT_PASSWORD: 'true'
     volumes:
-      - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
-      - ./web:/srv/api-platform/web
+      - 'db_data:/var/lib/mysql'
+    healthcheck:
+      test: ['CMD-SHELL', 'mysql --user=api_platform --password=api_platform -e "SHOW TABLES" api_platform']
```
teohhanhui: Please base this on https://github.com/docker-library/healthcheck/blob/master/mysql/docker-healthcheck

Also, the command(s) shouldn't be inlined in the Docker Compose file. :smile:

The other options (`interval`, `timeout`, `retries`) should be left to their defaults. Let the user customize those as they see fit.

docker-compose.yml
teohhanhui: Actually, the single volume mount only makes sense for development. Since we don't have separate distribution files for dev and prod, I prefer that we keep it as is to serve as guidance.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
+      timeout: '1s'
+      retries: 10
+
   php:
-    build: .
+    build: '.'
     depends_on:
-      - db
+      db:
+        condition: 'service_healthy'
```
teohhanhui: This is not supported in Docker Compose v3 file format, and thus should be avoided.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
+      - './:/srv/api-platform'
+
+  nginx:
+    image: 'nginx:1.11-alpine'
+    depends_on:
+      php:
+        condition: 'service_started'
+    ports:
+      - '80:80'
+    volumes_from:
```
teohhanhui: `volumes_from` is not supported in Docker Compose v3 file format, so it should not be used.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
-      - ./var:/srv/api-platform/var
-      - ./vendor:/srv/api-platform/vendor
-      - ./web:/srv/api-platform/web
+      - './:/srv/api-platform'
+
+  nginx:
+    image: 'nginx:1.11-alpine'
+    depends_on:
+      php:
+        condition: 'service_started'
```
teohhanhui: This is not supported in Docker Compose v3 file format, and thus should be avoided.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
-version: '2'
+version: '2.1'
+
+volumes:
```
rawkode: It is a norm in example, sure. But we should challenge this, why would we use a volume before it's defined? I think it's clearer to expose volumes up-front.

dunglas: As a general politic, we follow upstream conventions (for Symfony, Docker, Doctrine and so on). If it can be challenged, it should be in the upstream project.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
+  db_data:
```
rawkode: Because most editors (Vim, Atom and IntelliJ) don't highlight keys with dashes correctly

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
+    driver: 'local'
```
rawkode: I like to be explicit, happy to change

dunglas: I like too, but @teohhanhui is right, in we try to make default configs as short as possible to not confuse the end user.

docker-compose.yml
```diff
@@ -1,38 +1,44 @@
...
 
 services:
   db:
+    image: 'mysql:5.7'
```
rawkode: If you want the keys sorted that way, sure. This isn't the norm though.

Over-quoting is to remove confusion from newcomers to YAML who don't understand how crazy the semantics are and don't understand why we'd have:

```yaml
key: "True"
key2: This is ok though
```

docker-compose.yml
rawkode: I was refraining from adding another file, but sure.

docker-compose.yml
rawkode: `v3` isn't stable yet. It's also for SwarmMode, and most people using this file will be using it for development. You shouldn't restrict yourself by adopting v3 at this point in time.

docker-compose.yml
rawkode: Single volume mount makes sense for development, which is what this file is intended for. See above for v3 remarks

docker-compose.yml
rawkode: See above

docker-compose.yml
rawkode: It's also worth pointing out, these healthchecks wouldn't actually work until my PR was merged and released, which I believe was only released yesterday. You should assume that everyone is still on an older release for a while.

https://github.com/docker/docker/pull/31491

docker-compose.yml
teohhanhui: > If you want the keys sorted that way, sure. This isn't the norm though.

Yeah, I don't think there's any norm when it comes to the ordering of keys, and in the absence of any order-sensitive items (are there any?), alphabetical order seems the most sensible. Previously the services are also sorted alphabetically by name, please follow that.

teohhanhui: > Over-quoting is to remove confusion from newcomers to YAML who don't understand how crazy the semantics are

I understand, but as @dunglas said we try to follow the conventions of official sources. The merits of YAML is out of scope of this discussion.

docker-compose.yml
teohhanhui: v3 has been stable for a while now. v3.2 is the latest (stable) file format. Whether we like Swarm Mode or not, that's a separate discussion that we do not need to be having here. By avoiding what's not supported in v3, we ensure forward compatibility.

docker-compose.yml
teohhanhui: It'd be great if the editors can have better support then... (That affects all YAML, right? So I'd think it's an important feature to support, considering the popularity of YAML.)

I prefer if we stick to conventions from official sources.

docker-compose.yml
teohhanhui: > why would we use a volume before it's defined?

`docker-compose up` (or `docker stack deploy`) takes care of the order of creation, so we really don't need to concern ourselves with that. I think the main argument for putting `services` on top is because it's the "main" top-level key. The other top-level keys (e.g. `volumes`, `networks`) define things used by the `services`.

rawkode:
I'll suggest you simply revert my changes. I was trying to help, I don't really want to waste time debating non-existent / undocumented standards within your project.

docker-compose.yml
teohhanhui: Hmm... That's weird because a similar healthcheck script utilizing environment variables has been working for me: https://github.com/docker-library/healthcheck/blob/master/postgres/docker-healthcheck

dunglas:
@rawkode your contribution is really appreciated, don't worry! We'll keep your PR, we may made some changes to keep in sync with our CS.

dunglas:
@teohhanhui would you mind making a follow up PR with your suggested changes?

docker-compose.yml
dunglas: I just tried, and at least IntelliJ handle dashes correctly.

PR #258, opened 2017-03-21
Don't use ClassLoader component deprecated in Symfony 3.3

teohhanhui:
No need for conditionals because we require PHP >= 7.0


dunglas:
Thanks @teohhanhui

PR #281, opened 2017-04-11
Improve PHP container

luispabon:
Hey there, I'm the author of PHPDocker.io. I've been using API Platform for a number of projects and always been bothered by the PHP container using the official image. The official image isn't great, lacks a great many php extensions, and building everything from scratch bloats build times substantially. The tradeoff is a bigger image file at the end. The alpine image comes with its own problems around name resolution, especially when running within Kubernetes.

This PR refactors the  PHP container to use PHPDocker.io's and substantially reduce docker bootstrap times as no more extension compilation is required. PHP image is also half as big as a result. Cleanup everything else around it that doesn't need to be done anymore.


teohhanhui:
I think we should stick to the official image. If it's not good enough, it can be improved.

soyuka:
@dunglas it's your call but I agree with @teohhanhui on this. We can maintain our own image and if there's an issue with it we can fix it (instead of relying on a third-party docker image).

luispabon:
I went down the route of providing my own images on phpdocker.io for certain components (php and redis mostly) due to ocassional breakages happening upstream. I would recommend doing the same, and would be happy to help setting this up if you guys would be up for it. You'd need a repo and an account on docker hub.

teohhanhui:
I'd recommend against maintaining our own image. There's a reason why there are official Docker images.

soyuka:
Bad choice of word from me, I meant maintaining our own docker configuration in this repository in enough.

teohhanhui:
Oh yeah, actually we're building our own images. Just not pre-built and pushed to a repository. :laughing:

dunglas:
Those changes look interesting, but as pointed out by @teohhanhui and @soyuka, why not proposing those changes to the upstream official Docker image?

Simperfit:
We should stick with the official image as a source, but we should be using a pre-build image based on the current docker configuration.

teohhanhui:
Closing this as we are not going to move away from the official PHP image.

teohhanhui:
@Simperfit Sure, we could set up automated build and push for our Docker image. That should make it much faster to start a new project.

PR #289, opened 2017-04-22
Upgrade to Docker Compose v3

dunglas:
Also fix some issues raised by @teohhanhui in #244.

Ping @rawkode @jockos

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,9 +2,7 @@ version: '3'
 
 services:
   db:
-    image: mysql:8.0
-    # Temporary fix, see https://github.com/docker-library/mysql/issues/284
-    entrypoint: ['/entrypoint.sh', '--character-set-server=utf8', '--collation-server=utf8_general_ci']
+    image: mysql:5.7
     environment:
       MYSQL_DATABASE: api_platform
       MYSQL_USER: api_platform
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,44 +1,45 @@
-version: '2.1'
-
-volumes:
-  db_data:
-    driver: 'local'
+version: '3'
 
 services:
   db:
-    image: 'mysql:5.7'
+    image: mysql:8.0
+    # Temporary fix, see https://github.com/docker-library/mysql/issues/284
+    entrypoint: ['/entrypoint.sh', '--character-set-server=utf8', '--collation-server=utf8_general_ci']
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,44 +1,43 @@
...
-        condition: 'service_healthy'
+      - db
     environment:
-      SYMFONY_ENV: 'dev'
+      SYMFONY_ENV: dev
     volumes:
-      - './:/srv/api-platform'
+      - ./:/srv/api-platform
+      # For production only:
+      #- ./var/cache:/srv/api-platform/var/cache
```
Simperfit: The cache should be a non-shared volume even on dev because of performance issue on mac os x and windows

docker-compose.yml
```diff
@@ -1,44 +1,43 @@
...
+      #- ./var/logs:/srv/api-platform/var/logs
+      #- ./var/sessions:/srv/api-platform/var/sessions
+      #- ./var:/srv/api-platform/var
+      #- ./vendor:/srv/api-platform/vendor
```
Simperfit: The vendor should be a non-shared volume too IMO

docker-compose.yml
```diff
@@ -1,44 +1,43 @@
...
-      - 'db_data:/var/lib/mysql'
+      - db-data:/var/lib/mysql
     healthcheck:
-      test: ['CMD-SHELL', 'mysql --user=api_platform --password=api_platform -e "SHOW TABLES" api_platform']
-      interval: '5s'
-      timeout: '1s'
-      retries: 10
+      test: mysql --user=api_platform --password=api_platform -e "SELECT 1" api_platform
 
   php:
```
meyerbaptiste: [Like @teohhanhui](https://github.com/api-platform/api-platform/pull/248#issuecomment-284959726), I'm in favor of changing the name to `app`.

meyerbaptiste:
You have to upgrade Docker Engine to v1.13 in Travis: https://gist.github.com/dylanscott/ea6cff4900c50f4e85a58c01477e9473

docker-compose.yml
teohhanhui: Yes, I suggest we leave a comment for the user to comment out these sub-directory volumes if they are running on Linux, where there are no such performance issues.

dunglas:
Closed in favor of #308

Issue #298, opened 2017-05-12
Installing Docker version fails


bnd170:
When I execute `docker-compose up -d` I get this error:

> In file included from /usr/local/include/php/main/php_network.h:126:0,
                 from /usr/local/include/php/ext/standard/file.h:26,
                 from /usr/src/php/ext/zip/php_zip.c:27:
/usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
 #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
  ^
In file included from /usr/src/php/ext/zip/php_zip.c:29:0:
/usr/local/include/php/ext/pcre/php_pcre.h:29:18: fatal error: pcre.h: No such file or directory
compilation terminated.
Makefile:194: recipe for target 'php_zip.lo' failed
make: *** [php_zip.lo] Error 1
ERROR: Service 'php' failed to build: The command '/bin/sh -c set -xe 	&& apk add --no-cache --virtual .build-deps 		$PHPIZE_DEPS 		icu-dev 		zlib-dev 	&& docker-php-ext-install 		intl 		pdo_mysql 		zip 	&& pecl install 		apcu-${APCU_VERSION} 	&& docker-php-ext-enable --ini-name 20-apcu.ini apcu 	&& docker-php-ext-enable --ini-name 05-opcache.ini opcache 	&& apk del .build-deps' returned a non-zero code: 2

I'm using https://github.com/codekitchen/dinghy to increase performe of docker.

glanchow:
I ran the same issue, please add pcre-dev in the Dockerfile line 14 like this
`
	$PHPIZE_DEPS \
        pcre-dev \
	icu-dev \
`
I will submit a PR in a few minutes

teohhanhui:
This is not right. `$PHPIZE_DEPS` already includes `pcre-dev`: https://github.com/docker-library/php/blob/fba839b7dfbde029db36e834570a6b546189ee9e/7.1/fpm/alpine/Dockerfile

You may be using an outdated base image.

teohhanhui:
Okay, this was only recently broken since docker-library/php#429 and fixed by docker-library/php#433, so we can indeed revert now. Sorry for my lack of investigation...

bnd170:
Thanks guys! it worked.

PR #299, opened 2017-05-12
Add PCRE dependency to the docker build for the php image.

glanchow:
#298


Simperfit:
I think this PR should target 2.0 branch, but after that 👍

dunglas:
Thanks @glanchow

PR #302, opened 2017-05-15
Revert "Add PCRE dependency to the docker build for the php image."

teohhanhui:
Reverts api-platform/api-platform#299

See https://github.com/api-platform/api-platform/issues/298#issuecomment-301399836 and https://github.com/api-platform/api-platform/issues/298#issuecomment-301401182


dunglas:
Thanks @teohhanhui

PR #308, opened 2017-05-23
Revert #224; Upgrade to Docker Compose v3

teohhanhui:
This reverts commit e7e1ed6b87920ba21e923d797139a5f02f62d387.

Replaces #289.


dunglas:
What about #306 instead?

teohhanhui:
I suggest we should revert first, then we can see what we can improve from there.

dunglas:
It will be more work ;) Can you review #289 instead, I'll add missing parts :) We need this for the upcoming beta of 2.1!

docker-compose.yml
```diff
@@ -1,44 +1,52 @@
...
-
-  php:
-    build: '.'
-    depends_on:
-      db:
-        condition: 'service_healthy'
+      # Comment out the next line in production
+      - ./:/srv/api-platform:rw
+      # This is for assets:install
+      - ./web:/srv/tripviss-api/web:rw
```
dunglas: `tripviss-api` should probably be removed :p

docker/db/docker-healthcheck.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
dunglas: This file seems to come from https://github.com/docker-library/healthcheck/blob/master/mysql/docker-healthcheck
Can you add a comment with the original URL and copyright older?

```
# Copyright (c) 2014-2016 Docker, Inc.
# Origin: https://github.com/docker-library/healthcheck/blob/master/mysql/docker-healthcheck
# Licence: MIT
```

docker-compose.yml
teohhanhui: Oops... :smile:

.travis.yml
```diff
@@ -1,19 +1,22 @@
...
+  - sudo apt-get -y install "docker-ce=${DOCKER_VERSION}~ce-0~ubuntu-$(lsb_release -cs)"
+  - sudo sh -c 'curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m) > /usr/local/bin/docker-compose'
+  - sudo chmod +x /usr/local/bin/docker-compose
 
 script:
   - docker-compose build
-  - docker-compose run --rm php composer install --prefer-dist --no-progress --no-suggest --no-interaction
-  - docker-compose run --rm php bin/console security:check
-
+  - docker-compose run --rm app bin/console security:check
```
dunglas: Can you add:

```
  - docker-compose run --rm app bin/console doctrine:schema:create
```

app/config/config.yml
```diff
@@ -41,13 +41,14 @@ twig:
 # Doctrine Configuration
 doctrine:
     dbal:
-        driver: pdo_mysql
+        driver: 'pdo_mysql'
```
dunglas: Should be reverted (https://github.com/symfony/symfony-standard/blob/3.3/app/config/config.yml#L44), only strings containing special chars are quoted.

app/config/config.yml
```diff
@@ -41,13 +41,14 @@ twig:
...
         host: '%database_host%'
         port: '%database_port%'
         dbname: '%database_name%'
         user: '%database_user%'
         password: '%database_password%'
-        charset: UTF8
+        server_version: '5.7'
+        charset: 'UTF8'
```
dunglas: Same here

.travis.yml
teohhanhui: We'd have to run the `db` service for this to work, no? Maybe we can run this with `--dump-sql`?

app/config/config.yml
teohhanhui: Uhh... I thought it was single-quote-everything :laughing:

.travis.yml
dunglas: Indeed, the idea was to test that the Docker setup for the DB was properly running.

dunglas:
Thank you very much @teohhanhui

PR #313, opened 2017-05-29
Fix Nginx's config

dunglas:
Fixes #308.

### DIFF
--- a/docker/nginx/conf.d/default.conf
+++ b/docker/nginx/conf.d/default.conf
@@ -1,5 +1,6 @@
 server {
     root /srv/api-platform/web;
+    resolver 127.0.0.11;
 
     location / {
         # try to serve file directly, fallback to app.php
@@ -49,7 +50,4 @@ server {
     location ~ \.php$ {
       return 404;
     }
-
-    error_log /var/log/nginx/app_error.log;
-    access_log /var/log/nginx/app_access.log;
 }


docker/nginx/conf.d/default.conf
```diff
@@ -49,7 +50,4 @@ server {
     location ~ \.php$ {
       return 404;
     }
-
-    error_log /var/log/nginx/app_error.log;
```
Simperfit: why removing the logs ?

dunglas: I don't remove them, I keep the default location. It allows to have it both in files and on stdout/stderr thanks to a trick in the official Nginx image.

docker-compose.yml
```diff
@@ -39,6 +39,8 @@ services:
     build:
       context: ./docker/nginx
       dockerfile: ./Dockerfile
+    depends_on:
+      - app
```
teohhanhui: This is unnecessary. That's why I removed it.

docker/nginx/conf.d/default.conf
teohhanhui: > It allows to have them both in files and in stdout/stderr thanks to a trick in the official Nginx image.

That's not true. They are symlinked to stdout and stderr respectively.

docker/nginx/conf.d/default.conf
```diff
@@ -1,5 +1,6 @@
 server {
     root /srv/api-platform/web;
+    resolver 127.0.0.11;
```
teohhanhui: Good catch! I should have actually tried running it!

docker/nginx/conf.d/default.conf
dunglas: It was was I was calling a trick.

PR #314, opened 2017-05-30
Switch to Symfony 3.3 stable

dunglas:


composer.json
```diff
@@ -30,7 +29,8 @@
         "sensio/distribution-bundle": "^5.0.19",
         "sensio/framework-extra-bundle": "^3.0.2",
         "symfony/monolog-bundle": "^3.1.0",
-        "symfony/swiftmailer-bundle": "^2.3",
+        "symfony/swiftmailer-bundle": "^2.3.10",
+        "symfony/symfony": "3.3.*",
```
teohhanhui: Can't this be `^3.3`?

docker-compose.yml
```diff
@@ -16,11 +16,11 @@ services:
       # This is for assets:install
       - ./web:/srv/api-platform/web:rw
       # If you develop on Linux, comment out the following volumes to just use bind-mounted project directory from host
-      - /srv/api-platform/var
-      - /srv/api-platform/var/cache
-      - /srv/api-platform/var/logs
-      - /srv/api-platform/var/sessions
-      - /srv/api-platform/vendor
+      #- /srv/api-platform/var
```
teohhanhui: Why this change? Doesn't it cause performance problems on Mac OS, at least?

docker-compose.yml
```diff
@@ -45,7 +45,7 @@ services:
     volumes:
       # Comment out the next line in production
       - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
-      - ./web:/srv/api-platform/web:ro
+      - ./web:/srv/api-platform/web:rw
```
teohhanhui: Why this change? `nginx` has no business writing anything into this directory...

docker-compose.yml
Simperfit: it does, this should be reverted

composer.json
dunglas: I've kept this as in `symfony/symfony-standard` on purpose (it prevents unwanted upgrades to the next major version when it is released)

PR #325, opened 2017-06-07
Introduce a new file to register filters

dunglas:
It prevents to have to define manually `autoconfigure`, `autowire` and other keys because of the new `_defaults` key introduced in the standard edition of Symfony 3.3.

### DIFF
--- a/app/config/api_filters.yml
+++ b/app/config/api_filters.yml
@@ -1,0 +1,8 @@
+# Learn more about filters
+# https://api-platform.com/docs/core/filters
+services:
+    # configure filters
+    #foo.search_filter:
+    #    parent: 'api_platform.doctrine.orm.search_filter'
+    #    arguments: [ { id: 'exact', bar: 'partial' } ]
+    #    tags: ['api_platform.filter']
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -2,6 +2,7 @@ imports:
     - { resource: parameters.yml }
     - { resource: security.yml }
     - { resource: services.yml }
+    - { resource: api_filters.yml }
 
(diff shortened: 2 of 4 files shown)


app/config/config.yml
```diff
@@ -2,6 +2,7 @@ imports:
     - { resource: parameters.yml }
     - { resource: security.yml }
     - { resource: services.yml }
+    - { resource: filters.yml }
```
teohhanhui: `api_filters.yml`?

dunglas: I propose to keep it simple, the user can always rename it if he wants to.

teohhanhui: I think it's clearer to the user what are API filters, as opposed to just filters (which could be Twig filters or whatever...)

app/config/api_filters.yml
```diff
@@ -0,0 +1,8 @@
+# Learn more about filters
+# https://api-platform.com/docs/core/filters
+services:
+    # configure filters
+    #foo.search:
```
teohhanhui: I know this is a shortcut, but I think naming services like this is not sensible.

It'd be better if we name the service properly, at least `foo.search_filter` would be better. (I'd prefer a more verbose and properly namespaced `app.resource.foo.search_filter`, but that's just me.)

app/config/api_filters.yml
```diff
@@ -0,0 +1,8 @@
...
+    #    parent: 'api_platform.doctrine.orm.search_filter'
+    #    arguments: [ { id: 'exact', bar: 'partial' } ]
+    #    tags: [ { name: 'api_platform.filter' } ]
```
teohhanhui:
```yaml
    tags: [api_platform.filter]
```

would do.

PR #331, opened 2017-06-12
Switch to Symfony Flex

dunglas:
This is a work in progress.
Installation instructions (not really straightforward right now):

```
composer create-project api-platform/api-platform:dev-flex demo
cd demo
make serve # Temporary workaround to create the var/ directory...
docker-compose up
```


soyuka:
Work in progress or working progress? :laughing:

dunglas:
Fixed 💯

soyuka:
Did you removed all `md` files on purpose? LICENCE has been deleted too.

dunglas:
I'm not sure about the approach, maybe it would be better to version all files generated by Flex as well. This way it would not degrade the current developper experience while allowing to use Flex.

dunglas:
Yes, the spirit of Flex is to have a minimal repository (like `symfony/skeleton`) that the user can use without removing anything. But maybe can we start with a heavier setup as API Platform can already be installed easily with the minimal Flex :)

We can have 2 entrypoints:

* Battery-included: download `api-platform/api-platform` and type `docker-compose up`
* Minimal: `composer create-projet symfony/skeleton app && cd app && composer req api`

WDYT?

soyuka:
`composer req api` would use which repository? 

Okay, I get the skeleton part now and I think it's a good idea for advanced users that want to set up an api in no time! Though, for newbies, having a repository to clone and then to issue `docker-compose up` would probably be ~easier~ less scarier.

dunglas:
`api` is an official Flex alias of `api-platform/api-pack`: https://github.com/api-platform/api-pack

dunglas:
Closing in favor of #422.

PR #333, opened 2017-06-16
Minor docker improvements

akira28:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ RUN apk add --no-cache --virtual .persistent-deps \
 		icu-libs \
 		zlib
 
+ENV APCU_VERSION 5.1.8
+
 RUN set -xe \
 	&& apk add --no-cache --virtual .build-deps \
 		$PHPIZE_DEPS \
@@ -14,7 +16,8 @@ RUN set -xe \
 		intl \
 		pdo_mysql \
 		zip \
-	&& pecl install apcu \
+	&& pecl install \
+		apcu-${APCU_VERSION} \
 	&& docker-php-ext-enable --ini-name 20-apcu.ini apcu \
 	&& docker-php-ext-enable --ini-name 05-opcache.ini opcache \
 	&& apk del .build-deps
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,6 @@ RUN apk add --no-cache --virtual .persistent-deps \
 		icu-libs \
 		zlib
 
-ENV APCU_VERSION 5.1.8
-
 RUN set -xe \
 	&& apk add --no-cache --virtual .build-deps \
 		$PHPIZE_DEPS \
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.2'
```
dunglas: What's the benefit of this change?

Dockerfile
```diff
@@ -5,8 +5,6 @@ RUN apk add --no-cache --virtual .persistent-deps \
 		icu-libs \
 		zlib
 
-ENV APCU_VERSION 5.1.8
```
Simperfit: I think we should be keeping specify the version.

docker-compose.yml
Simperfit: I agree it does not seem to change anything.

docker/nginx/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM nginx:1.11-alpine
+FROM nginx:mainline-alpine
```
Simperfit: Why ?

docker-compose.yml
akira28: Doesn't change anything (at the moment), is just "newer is better" stuff 💃

Dockerfile
akira28: Why? Isn't it better to have the latest stable version for performance and security, instead of having to change it every time a new version is out?

docker/nginx/Dockerfile
akira28: Same reason, performance, security and features are available without having to change this file each time a new version is released.

Dockerfile
Simperfit: Because it can break things.

docker/nginx/Dockerfile
Simperfit: same reason

akira28:
I've restored the apcu fixed version, and set nginx version to 1.13

dunglas:
Looks good to me (but I still don't get the point of specifying the minor version of the docker compose format). Waiting for @teohhanhui's approval before merging.

akira28:
👍
Also, I would gladly help with #334 , if you are interested

Simperfit:
@akira28 We will be doing it, we just need to be sure that everything is ok before doing it :).

akira28:
Of course 👌

dunglas:
Can you rebase please @akira28? Can you also update the `version` field to `3.3`?

dunglas:
After double checking, most examples on docker.com use `versino: '3'`. We should continue using this.

teohhanhui:
Let's just open another PR to bump the nginx version.

PR #351, opened 2017-07-02
Fix Travis

dunglas:


teohhanhui:
Why?

teohhanhui:
How do we make sure that we're using the correct version of Docker and Docker Compose going forward?

dunglas:
Travis has upgraded its Ubuntu images and our test suite was broken. Removing specific versions actually fixes it. We should let Travis do its best when possible (here, its possible). I suggest to keep it as is to prevent breakages when Travis upgrade its setup.

dunglas:
See https://travis-ci.org/api-platform/api-platform/builds/249307547?utm_source=github_status&utm_medium=notification for instance.

teohhanhui:
Make sense...

PR #355, opened 2017-07-06
Dockerfile : Adding required lib pcre

sylvaincombes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

When I tried `docker-compose up -d` like indicated in the documentation the build fails  with a message like "missing pcre.h ..."
I found that in order to the docker build to succeed we need the libpcre3-dev package in the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN buildDeps=" \
     && apt-get update \
     && apt-get install -y --no-install-recommends \
         $buildDeps \
+        libpcre3-dev \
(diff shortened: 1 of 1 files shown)


teohhanhui:
Duplicate of #299

Should already have been fixed by docker-library/php#433

PR #398, opened 2017-08-22
Fix dependency conflict

### LINKED ISSUE #397: Dependencies problem
Hi,

When using api-platform 2.1, we cannot upgrade `symfony/symfony` to 3.3.5 (or upper) due to conflict with `phpdocumentor/reflection-docblock 3.2`.

```
$ composer why-not symfony/symfony 3.3.6
symfony/symfony  v3.3.6  conflicts  phpdocumentor/reflection-docblock (<3.0||>=3.2.0)  
```

Maybe we can upgrade phpdocumentor/reflection-docblock to 4.0 but what is the consequence for api-platform ?

erichard:
Downgrade phpdocumentor/reflection-docblock to 3.1.x to be able to use latest symfony/symfony.

phpdocumentor/reflection-docblock 4.x is also set to conflict in the symfony composer.json

Fix #397

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "guzzlehttp/guzzle": "^6.0",
         "incenteev/composer-parameter-handler": "^2.0",
         "nelmio/cors-bundle": "^1.4",
-        "phpdocumentor/reflection-docblock": "^3.0",
+        "phpdocumentor/reflection-docblock": "3.1.*",
         "sensio/distribution-bundle": "^5.0.19",
         "sensio/framework-extra-bundle": "^3.0.2",
         "symfony/monolog-bundle": "^3.1.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "2afb10084dda1a6724b080a89f72dc5e",
+    "content-hash": "8aed7dc345658bdbcbd603031b8c28e0",
     "packages": [
         {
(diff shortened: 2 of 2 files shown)


erichard:
@dunglas Tests are green

composer.json
```diff
@@ -26,7 +26,7 @@
         "guzzlehttp/guzzle": "^6.0",
         "incenteev/composer-parameter-handler": "^2.0",
         "nelmio/cors-bundle": "^1.4",
-        "phpdocumentor/reflection-docblock": "^3.0",
+        "phpdocumentor/reflection-docblock": "3.1.*",
```
dunglas: I don't get what it fixes. 3.1 is allowed by `^3.0`

erichard: Yes but 3.2 is allowed to. symfony composer.json conflict with 
```"phpdocumentor/reflection-docblock": "<3.0||>=3.2.0,<3.2.2",```

I don't know how composer choose what version to install in this case as there is multiple valid choices:
 *  symfony/symfony@3.3.4 with phpdocumentor/reflection-docblock@3.2.2 
 * symfony/symfony@3.3.6 with phpdocumentor/reflection-docblock@3.1.1

This is not really a bug in api-platform. I can downgrade phpdocumentor/reflection-docblock in my own project and it will also work.

Actually symfony/symfony is locked at 3.3.5 in the master branch, but if you try a `composer update` it will downgrade to 3.3.4.

meyerbaptiste: So, `"^3.0,<3.2"`?

dunglas:
This has been fixed in Symfony: https://github.com/symfony/symfony/pull/23848

Using `3.3@dev` until a new patch version of Symfony is released should do the trick. Closing for now.

PR #422, opened 2017-09-20
Revamp the distribution: Flex, JavaScript containers, HTTP/2 support, Helm chart

dunglas:
**Work in progress**

* Rethink the whole design of the distribution
* Rewrite the API skeleton to use Flex (`php`)
* Add a container for the client generator (`client`) - http://localhost:3000 / https://localhost:8000
* Add a container for the admin (`admin`) - http://localhost:3001 / https://localhost:8001
* Add a HTTP/2 reverse proxy with TLS support (self-signed certificate for dev ony, `h2-proxy`) - https://localhost
* Still, invalidation cache mechanism using Varnish (`cache-proxy`)

Test it:

```
git clone git@github.com:dunglas/api-platform.git
cd api-platform
git checkout flex-js
docker-compose up
```

TODO:

* [x] Add a Helm chart
* [x] Add a welcome page


soyuka:
Question, why do we need a container and an http server for `client-generator`?

admin/src/App.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from 'react';
+import { HydraAdmin } from '@api-platform/admin';
+
+class App extends Component {
+  render() {
+    console.log(process.env.REACT_APP_API_ENTRYPOINT);
```
soyuka: forgot to remove?

api/docker/varnish/conf/default.vcl
```diff
@@ -54,11 +54,6 @@ sub vcl_recv {
 
     return(synth(400, "ApiPlatform-Ban-Regex HTTP header must be set."));
   }
-
-  # Don't cache in dev mode
-  if (req.url ~ "^/app_dev.php") {
-    return(pass);
-  }
```
soyuka: not needed anymore?

api/docker/varnish/conf/default.vcl
```diff
@@ -3,22 +3,22 @@ vcl 4.0;
...
-    .window = 5;
-    .threshold = 3;
-  }
+  #.probe = {
+  #  .url = "/";
+  #  .timeout = 5s;
+  #  .interval = 10s;
+  #  .window = 5;
+  #  .threshold = 3;
+  #}
```
soyuka: dead code?

admin/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:8.5-alpine
+
+# Prevent the reinstallation of node modules at every changes in the source code
+RUN mkdir -p /usr/src/admin /tmp/deps
```
stof: Why not doing the `yarn install` in `/usr/src/admin` directly rather than doing it in tmp and then moving it later ?

api/docker/varnish/conf/default.vcl
Lctrs: I proposed a PR to bring this back if wanted. See https://github.com/dunglas/api-platform/pull/1.

api/.env
```diff
@@ -0,0 +1,21 @@
+# This file is a "template" of which env vars needs to be defined in your configuration or in a .env file
```
stof: if it is a template, shouldn't it be `.env.dist` ?

api/composer.json
```diff
@@ -0,0 +1,52 @@
+{
+    "type": "project",
+    "license": "proprietary",
```
stof: is it actually proprietary ? This looks weird to me

api/docker/varnish/conf/default.vcl
dunglas: It deserves a comment, remove the `#`s to enable probe (probably not needed in a managed environment like Kubernetes though).

api/docker/varnish/conf/default.vcl
dunglas: `app_dev.php` has been removed. Now both the nginx and the varnish containers are publicly exposed, type `https://localhost` for the non-cached version and `http://localhost:81` for the proxied version.

admin/Dockerfile
dunglas: To take advantage of Docker cache layers. This way the cache is used if dependencies have not been modified even if the source code code has been modified.

api/.env
dunglas: This line must be removed, good catch. Unlike Flex, I decided to follow the Create React App strategy (https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-development-environment-variables-in-env), so the `.env` file must be versioned.

I'm not sure if it's the best way to do anyway.

api/composer.json
dunglas: This repo is MIT, but the project created by the user may be proprietary. It's the file generated by Flex, I've not updated it. Maybe should I just remove this line.

admin/Dockerfile
stof: you could build them in the right folder and still use the Docker cache (you would even make the cache more efficient as you will avoid the layer moving files around)

I'm not asking you to stop install deps before adding the project code. I'm asking you to add the package.json and yarn.lock files directly in `/usr/src/admin` and do the install there, removing the need to move files around

dunglas: Fixed thanks. I've done the same thing for Composer btw.

dunglas:
[Helm](https://helm.sh/) chart added in 7cb72967c01853395aa3dd43b34ef88ce19436a9.

api/Dockerfile
```diff
@@ -40,29 +41,22 @@ ENV COMPOSER_ALLOW_SUPERUSER 1
...
-COPY bin bin/
-COPY src src/
-COPY web web/
+COPY . ./
 
-RUN composer dump-autoload --optimize --classmap-authoritative --no-dev
+RUN mkdir -p var/cache var/logs var/sessions \
+    && composer dump-autoload --optimize --classmap-authoritative --no-dev \
+    # Permissions hack because setfacl does not work on Mac and Windows
+	&& chown -R www-data var
```
greg0ire: This looks like a tabulation

api/Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM php:7.1-fpm-alpine
 RUN apk add --no-cache --virtual .persistent-deps \
 		git \
 		icu-libs \
+		make \
+		postgresql-dev \
```
Lctrs: Container size can be reduced by moving `postgresql-dev` to build-deps and replacing it by `libpq` in persistent-deps.

api/Dockerfile.varnish
```diff
@@ -2,7 +2,8 @@ FROM alpine:3.5
```
Lctrs: Why not bumping it to `alpine:3.6` ?

api/Dockerfile
```diff
@@ -14,23 +16,22 @@ RUN set -xe \
...
-COPY docker/app/install-composer.sh /usr/local/bin/docker-app-install-composer
-RUN chmod +x /usr/local/bin/docker-app-install-composer
+COPY docker/php/install-composer.sh /usr/local/bin/install-composer
+RUN chmod +x /usr/local/bin/install-composer
 
 RUN set -xe \
-	&& apk add --no-cache --virtual .fetch-deps \
-		openssl \
-	&& docker-app-install-composer \
+	&& apk add --no-cache --virtual .fetch-deps openssl \
```
Lctrs: `openssl` is now installed as persistent deps in parent image. So this can be removed.

Saphyel:
You need to add the `public/.htaccess`

dunglas:
@Saphyel, I don't think so... the provided setup use Nginx.

PR #438, opened 2017-10-23
Improve the Docker setup

dunglas:
* Use multi-stage builds:
  * to generate the certificate
  * to copy composer
* Optimize the PHP image
* Upgrade Node version


api/Dockerfile
```diff
@@ -25,34 +26,27 @@ RUN set -xe \
...
-	&& composer dump-autoload --optimize --classmap-authoritative --no-dev \
-	&& chown -R www-data var # Permissions hack because setfacl does not work on Mac and Windows
-
-COPY docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
-RUN chmod +x /usr/local/bin/docker-entrypoint
-
-ENTRYPOINT ["docker-entrypoint"]
-CMD ["php-fpm"]
+	&& composer dump-autoload --classmap-authoritative --no-dev \
+	&& chown -R www-data var
```
Simperfit: why removing the comment ?

dunglas: Because it appears in the shell...

Simperfit: It's because it's in the RUN, it should be before the RUN to explain why we chown in here

dunglas: Yes but it's just about a single line of the run. Putting it before makes no sense. It's why I've deleted it.

PR #457, opened 2017-11-13
Use the API in HTTPS by default to prevent mixed content errors

dunglas:


api/composer.json
```diff
@@ -46,7 +46,7 @@
     },
     "extra": {
         "symfony": {
-            "id": "",
+            "id": "01BYTDAT5NWGHQFHWQZR5GH14G",
```
sroze: TBR

PR #517, opened 2018-01-12
Use scanelf to detect runtime deps

Lctrs:
Applying same method as upstream docker php image.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,11 +1,8 @@
 FROM composer:1.6
 FROM php:7.2-fpm-alpine3.7
 
-RUN apk add --no-cache --virtual .persistent-deps \
-		git \
-		icu-libs \
-		libpq \
-		zlib
+RUN apk add --no-cache \
+		git
 
 ENV APCU_VERSION 5.1.9
 RUN set -xe \
@@ -22,6 +19,13 @@ RUN set -xe \
 		apcu-${APCU_VERSION} \
 	&& docker-php-ext-enable --ini-name 20-apcu.ini apcu \
 	&& docker-php-ext-enable --ini-name 05-opcache.ini opcache \
+	&& runDeps="$( \
+		scanelf --needed --nobanner --format '%n#p' --recursive /usr/local/lib/php/extensions \
+			| tr ',' '\n' \
+			| sort -u \
+			| awk 'system("[ -e /usr/local/lib/" $1 " ]") == 0 { next } { print "so:" $1 }' \
+	)" \
+	&& apk add --no-cache --virtual .php-phpexts-rundeps $runDeps \
 	&& apk del .build-deps
 
 COPY --from=0 /usr/bin/composer /usr/bin/composer


api/Dockerfile
```diff
@@ -1,28 +1,40 @@
...
+	pecl install \
 		apcu-${APCU_VERSION} \
-	&& docker-php-ext-enable --ini-name 20-apcu.ini apcu \
-	&& docker-php-ext-enable --ini-name 05-opcache.ini opcache \
-	&& apk del .build-deps
+	; \
+	\
+	docker-php-ext-enable --ini-name 20-apcu.ini apcu; \
+	docker-php-ext-enable --ini-name 05-opcache.ini opcache; \
+	\
```
jewome62: Why add NL here ? (and the others lines)

api/Dockerfile
```diff
@@ -1,28 +1,40 @@
...
+	runDeps="$( \
+		scanelf --needed --nobanner --format '%n#p' --recursive /usr/local/lib/php/extensions \
```
jewome62: What is the advantage to do this ? can you explain the difference ?

api/Dockerfile
Lctrs: They're only cosmetic, to clearly differentiate the different parts of the process.

api/Dockerfile
Lctrs: Before, we specified both runtime and build deps manually.

With this method, runtime deps are detected automatically. This ensures we don't left any packages that are not necessary to runtime in the images. It's also possibly strip down a little the final container size.
Also, it ensures we don't screw up with runtime deps in the future, since related errors can mostly be detected at runtime, not during the container's build.

dunglas:
Can you remove the cosmetic changes please? It makes the diff harder to review (and the final result is harder to read, but that's just my opinion).

Lctrs:
@dunglas Done.

api/Dockerfile
```diff
@@ -1,28 +1,35 @@
...
-		libpq \
-		zlib
+RUN apk add --no-cache \
+		git
 
 ENV APCU_VERSION 5.1.8
-RUN set -xe \
-	&& apk add --no-cache --virtual .build-deps \
+RUN set -xe; \
+	apk add --no-cache --virtual .build-deps \
```
dunglas: Why replacing `;` by `&&`? If one command fail, the process must stop.

Lctrs: > If one command fail, the process must stop.

It's the case.
I just found it more readable with `;`.

dunglas: Can we keep with the current CS please :)?

api/Dockerfile
```diff
@@ -1,28 +1,35 @@
 FROM composer:1.6
 FROM php:7.2-fpm-alpine3.7
 
-RUN apk add --no-cache --virtual .persistent-deps \
-		git \
-		icu-libs \
```
dunglas: Why removing those libs?

Lctrs: Because we don't need to specify them anymore. They're installed with the dev packages.

If I would translate the workflow of this PR to a Debian-based one, it would sound like this : 
1. Install the dev packages
2. Compile the extensions
3. Scan the compiled extensions to automatically detect their runtime dependencies and mark the corresponding packages as manually installed so they won't be auto removed during the next step
4. Auto-remove all the packages that have not been marked as manually installed.

This PR aims to only define the build dependencies in the Dockerfile and to keep only the necessary libs in the resulting image.

Is this explanation clear enough ? :/

api/Dockerfile
Lctrs: Done. :)

api/Dockerfile
jewome62: Hello,
I understand the explanation.
but git can be used for composer

dunglas:
Thanks @Lctrs

PR #526, opened 2018-01-19
Upgrade Flex

dunglas:


api/templates/base.html.twig
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
jewome62: This is a useless file for ApiPlatform.
The templating dir is here because `symfony\template` is need for dev version (it should be for profiler)

Maybe replace by a `.gitkeep` file ?

api/Dockerfile
```diff
@@ -1,13 +1,13 @@
-FROM composer:1.5
-FROM php:7.1-fpm-alpine
+FROM composer:1.6
+FROM php:7.2-fpm-alpine
```
Lctrs: Why not `php:7.2-fpm-alpine3.7` ? Actually, `php:7.2-fpm-alpine` point to the Alpine 3.6 based images.

api/templates/base.html.twig
dunglas: I'm hesitating. I agree with you, this useless for API Platform projects. On the other hand, it allows to be in standard Flex directory.

PR #574, opened 2018-02-16
Fixed docker for Windows

pborreli:
This is necessary on Windows otherwise the following error occurs on docker-compose up

```
php_1          | Nothing to install or update
php_1          | Generating autoload files
php_1          | Executing script cache:clear [OK]
php_1          |
php_1          | Executing script assets:install [OK]
': No such file or directoryexecute 'php


dunglas:
Well, it's weird but thanks!

PR #582, opened 2018-02-23
Update admin

mauchede:
Current version of `@api-platform/admin` does not work. I fixed the problem in `v0.3.1` (see https://github.com/api-platform/admin/pull/81).


dunglas:
Thanks @mauchede

PR #608, opened 2018-03-17
Fix dockerfile alternate paths

bobey:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | - 
| License       | MIT
| Doc PR        | Didn't find impacted documentation

Remove ./ relative alternate dockerfile path notation which doesn't work on latest version of Docker + Compose

See: https://docs.docker.com/compose/compose-file/#dockerfile

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -20,7 +20,7 @@ services:
     api:
         build:
             context: ./api
-            dockerfile: ./Dockerfile.nginx
+            dockerfile: Dockerfile.nginx
         depends_on:
             - php
         ports:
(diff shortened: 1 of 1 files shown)


bobey:
Yes it seemed like a bug but I didn't find the related issue you mentioned. Anyway, it now reflects exactly the official docker-compose documentation.

Not relevant to the current conversation but let me thank you for the awesome work you all made on this project.

PR #613, opened 2018-03-19
Push Docker images to container registry

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A

@dunglas will have to create the `api-platform` organization on [quay.io](https://quay.io) and set the following environment variables in `api-platform/api-platform` repository on Travis CI:

Name                    | "Display value in build log" | Example
----------------------- | ---------------------------- | ---------------------
DOCKER_REGISTRY         | Y                            | quay.io
DOCKER_USERNAME         | N                            | dunglas@gmail.com
DOCKER_PASSWORD         | N                            | xxxxxxxx
CONTAINER_REGISTRY_BASE | Y                            | quay.io/api-platform

And create the following public repositories on [quay.io](https://quay.io):

- api-platform/php
- api-platform/nginx
- api-platform/varnish
- api-platform/client
- api-platform/admin

It should only push the images when the Travis build is run on the main repository (`api-platform/api-platform`) and only for the `master` branch.


.editorconfig
```diff
@@ -43,7 +43,15 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = false
 
-[Dockerfile]
+[.travis.yml]
+indent_style = space
+indent_size = 2
```
teohhanhui: Using 2 spaces following official examples.

.editorconfig
```diff
@@ -43,7 +43,15 @@ indent_style = space
...
+
+[docker-compose{,.override}.yml]
+indent_style = space
+indent_size = 2
```
teohhanhui: Using 2 spaces following official examples.

.editorconfig
```diff
@@ -43,7 +43,15 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = false
 
-[Dockerfile]
+[.travis.yml]
+indent_style = space
+indent_size = 2
+
+[docker-compose{,.override}.yml]
```
teohhanhui: Use the name `docker-compose.yml` which is more commonly referred to in official documentation and examples. (Yes, I know it says that `docker-compose.yaml` also works...)

dunglas: I changed this for consistency. I find it weird to have some .yml and some .yaml in the same repository. And .yaml is recommended in the YAML spec IIRC.

dunglas: http://www.yaml.org/faq.html

teohhanhui: If it's our own YAML file, sure. But this is for Docker Compose. We've decided to stick to upstream conventions, right?

dunglas: Upstream support both. I'm not sure :)

.env
```diff
@@ -0,0 +1 @@
+CONTAINER_REGISTRY_BASE=quay.io/api-platform
```
sroze: I'd suggest using Docker Hub instead. There are multiple reasons. The first one is that it's the default registry, therefore we can simply ask users to `docker pull api-platform/api-platform` if we have so. It's pretty much familiar to what they have already with PHP & composer. Also, Docker Hub images are cached at various network boundaries (for examples within Google's network) while Quay's ones are not.

.editorconfig
sroze: I'm not fond of this one. I'd prefer keeping 4 spaces for consistency.

.editorconfig
dunglas: After a second though, I think (like @sroze) we should name and indent all YAML files consistently. YAML is an open format, there is no point to use different flavors of it in the same project (ours). The benefit of using YAML over custom config formats is precisely that it's an open standard with parsers in many languages.
 
The only exception regarding the extension, is `.travis.yml`, because AFAIK `.travis.yaml` isn't supported :(

.travis.yml
```diff
@@ -31,3 +33,14 @@ script:
   - curl -k https://localhost:444 # Admin (HTTP/2)
   - curl -k https://localhost:8443 # API (HTTP/2)
   - curl -k https://localhost:8444 # Varnish (HTTP/2)
+
+before_deploy:
+  - echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin "$DOCKER_REGISTRY"
+
+deploy:
```
Simperfit: do we really need to use that for a single command line ?

.env
dunglas: A benefit of using Quay: our org name is still available (Docker Hub doesn't allow dashes and `apiplatform` is not available. I registered `apipltfrm` but it's less nice.

Simperfit:
@teohhanhui @dunglas @sroze I guess we can do this just by enabling continuous-pipe on api-platform. 
Since we already have a cluster at Les-Tilleuls.coop with some part of api-platform in it, I guess we could install continuous-pipe on that cluster, enable or not a feature-branch based on the patch made in the core or juste deploy and build a version automatically when pushing to master.

WDYT ?

dunglas:
@Simperfit sorry but -1 on my side. It means more maintenance, more ops and costs for the projects. I prefer to automate this fully using managed services.

.travis.yml
teohhanhui: Yes, we need to, for conditional deployment. Only on the main repository (api-platform/api-platform) and the master branch.

.editorconfig
teohhanhui: Well, of course we could do it either way but sticking to upstream conventions means it's easier to copy and paste from official examples. And to me consistency with other projects is always good. It means if you've worked with Docker Compose before you're already familiar with the spacing.

.env
teohhanhui: If we really need to switch to Docker Hub later on, it's a very easy switch anyway.

api/Dockerfile
```diff
@@ -57,7 +57,7 @@ COPY . ./
 RUN mkdir -p var/cache var/logs var/sessions \
 	&& composer dump-autoload --classmap-authoritative --no-dev \
 	&& composer run-script --no-dev post-install-cmd \
-	&& chmod +x bin/console \
+	&& chmod +x bin/console && sync \
```
teohhanhui: Workaround for https://github.com/moby/moby/issues/9547 :smiley:

.env
dunglas: Ok, let's use Quay for now.

api/Dockerfile
dunglas: Nice one!

docker-compose.yml
```diff
@@ -0,0 +1,112 @@
+version: '3.2'
```
dunglas: `git mv` to not loose the history?

.env
dunglas: May I suggest to hardcode `quay.io/api-platform`? So one less file in the repo, and it's easier to dig into for a newcomer. WDYT?

teohhanhui: So that we have to change it in multiple places instead of one, and it can't be overridden by an environment variable? I don't think that's a good idea.

docker-compose.yml
teohhanhui: I've just tried it. It makes no difference. Git calculates that dynamically, and `git log docker-compose.yml` is able to connect the dots just fine in this case (and GitHub fails to do the same, even if I put the rename into a separate commit). :smile:

api/Dockerfile
teohhanhui: This is just in case we decide to run some bin/console commands in the same `RUN`. :laughing:

dunglas:
Thanks ou @teohhanhui

PR #616, opened 2018-03-20
Update the structure to separate source-code away from the build files and to easily enabling git submodules

sharaf84:
Update the structure to separate source-code away from the build files and to easily enabling git submodules


teohhanhui:
I'm definitely -1 for this. What is the problem with using git submodules?

teohhanhui:
Closing as this doesn't really make sense.

PR #663, opened 2018-04-25
Minor enhancements for docker-compose.yml file

nayzo:
Update docker-compose.yml indentation and extension to match the same format used in Symfony 4


teohhanhui:
Sorry, we've been having back-and-forth discussions about this. And in fact, it's only recently been changed to follow upstream `docker-compose.yml`: https://docs.docker.com/compose/compose-file/

teohhanhui:
Please make sure you're using EditorConfig for your IDE / text editor. It shouldn't be a problem since we have an `.editorconfig` file.

NAYZO:
It's the case. Thanks for the clarification.

PR #666, opened 2018-05-03
Use a single Dockerfile for the API images

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | N/A
| License       | MIT
| Doc PR        | N/A?


api/docker/varnish/docker-entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
teohhanhui: I've adapted this from https://github.com/tripviss/docker-varnish

api/Dockerfile
```diff
@@ -66,3 +70,26 @@ RUN chmod +x /usr/local/bin/docker-entrypoint
...
 ENTRYPOINT ["docker-entrypoint"]
 CMD ["php-fpm"]
+
+FROM nginx:${NGINX_VERSION}-alpine AS api_platform_nginx
+
+COPY docker/nginx/conf.d /etc/nginx/conf.d/
+
+WORKDIR /srv/api
+
+COPY --from=api_platform_php /srv/api/public public/
```
teohhanhui: This means we copy installed assets too. So the nginx image is now more production ready. Yay! :smile:

api/docker/varnish/docker-entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+# first arg is `-f` or `--some-option`
+if [ "${1#-}" != "$1" ]; then
+	set -- varnishd "$@"
+fi
+
+# default args
+if [ "$1" = 'varnishd' ] && [ $# -eq 1 ]; then
+	set -- varnishd \
+		-F \
+		-a :"${VARNISH_PORT:-6081}" \
```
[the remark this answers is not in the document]
dunglas: I don't understand why we need to change that. It's a container with a single process, using the default HTTP port:

* looks obvious
* makes the `.env` file better
* allow to not use port mapping in our Kubernetes container

And anyway, cannot merge it as is because it will break this: https://github.com/api-platform/api-platform/blob/master/api/helm/api/templates/varnish-service.yaml#L13-L16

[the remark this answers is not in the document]
teohhanhui: I don't see what's the point of changing it to port 80 when the upstream default is 6081. And that's what people would be familiar with / expect it to be. It's just like port 3000 for an Express.js app. We could of course use port 80. But that'd go against the conventions.

But anyway, if it breaks but our build is still green, then we need to add more tests to the build? :)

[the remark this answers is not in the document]
dunglas: Unlike `client` and `admin`, this isn't (only) a development container: it can be used in production. The Varnish documentation I quoted explicitly state that `6081` is for testing purpose, I actually never seen this port used for Varnish in production.

This isn't against the conventions:

> HTTP communication usually takes place over TCP/IP connections. The default port is TCP 80

(HTTP 1.1 RFC)

Adding smoke tests for the Kubernetes as we do for Docker Compose would be great. We'll have to install Minikube in Travis for that.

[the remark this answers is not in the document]
dunglas: However, it would make sense to expose Varnish on host's `6081` port (because 80 isn't free). But it would be a BC break, and would require us to update the documentation, so I suggest to keep this as is.

api/Dockerfile
Simperfit: I don't think we should use --from, because it does not work with google cloud atm :/.

teohhanhui: What do you mean? The images are built by Docker. Kubernetes has nothing to do with it?

Simperfit: I mean if you are using Google Container Builder or you are building directly from Google it won't work.

teohhanhui: Uhh... But I don't think we should let that stop us from using the new features.

api/helm/api/values.yaml
```diff
@@ -40,14 +40,14 @@ ingress:
...
-  #url: pgsql://api-platform:!ChangeMe!@example.com/api?serverVersion=10.1
+  #url: pgsql://api-platform:!ChangeMe!@example.com/api
   postgresUser: api-platform
   postgresPassword: ChangeMe
   postgresDatabase: api
   # Persistent Volume Storage configuration.
   # ref: https://kubernetes.io/docs/user-guide/persistent-volumes
   persistence:
-    enabled: false
+    enabled: true
```
teohhanhui: A sensible default? Since the postgresql chart also defaults to persistence enabled.

api/helm/api/values.yaml
```diff
@@ -8,21 +8,21 @@ secret: ChangeMe
 corsAllowOrigin: http://example.com
 
 php:
-  repository: gcr.io/test-api-platform/php
+  repository: quay.io/api-platform/php
```
teohhanhui: This should make it easier to deploy since we have the built images in the repositories now.

api/helm/api/requirements.yaml
```diff
@@ -1,5 +1,5 @@
 dependencies:
 - name: postgresql
-  version: 0.8.1
+  version: 0.11.0
```
teohhanhui: @dunglas Is it a good idea to bump the version here?

api/helm/api/requirements.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
```
teohhanhui: @dunglas Is it a good idea to check this in? Why or why not?

api/Dockerfile
dunglas: Having full GCP support is a nice feature of API Platform (and we at @coopTilleuls use it on some project). I think we shouldn't switch to the new feature until it is supported by at least GCP.

WDYT @api-platform/core-team

api/helm/api/requirements.lock
dunglas: I don't know what is the good practice regarding this. I guess it's a good idea but I'm not sure.

api/Dockerfile
teohhanhui: But we already use it for composer?

api/helm/api/requirements.yaml
dunglas: Unfortunately we cannot, Postgres 11 is buggy with Doctrine (especially the `doctrine:schema:update` command).

teohhanhui: No, this is the version of the Helm chart, not the version of PostgreSQL. :smile:

api/helm/api/values.yaml
dunglas: The rationale was to strongly encourage to use the managed RDBMS provided by the cloud platform (Google Cloud SQL, Amazon RDS, Heroku Postgres...) instead. Please keep it as is (maintaining a RDBMS in Kubernetes is very hard).

api/helm/api/requirements.yaml
dunglas: Oops 🤣. It it works then 👍 !

api/Dockerfile
dunglas: Indeed, so if it wasn't working already, then 👍

api/Dockerfile
```diff
@@ -1,6 +1,8 @@
 ARG PHP_VERSION=7.2
+ARG NGINX_VERSION=1.13
 ARG ALPINE_VERSION=3.7
-FROM php:${PHP_VERSION}-fpm-alpine${ALPINE_VERSION}
+
+FROM php:${PHP_VERSION}-fpm-alpine AS api_platform_php
```
teohhanhui: We no longer need the `ALPINE_VERSION` here since https://github.com/docker-library/php/pull/650 :tada:

api/.env
```diff
@@ -13,7 +13,7 @@ TRUSTED_HOSTS=localhost,api
 # Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
 # For an SQLite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
 # Configure your db driver and server_version in config/packages/doctrine.yaml
-DATABASE_URL=pgsql://api-platform:!ChangeMe!@db/api
+DATABASE_URL=postgres://api-platform:!ChangeMe!@db/api
```
teohhanhui: `postgres://` protocol is our best bet.

https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING
https://devcenter.heroku.com/articles/heroku-postgresql

api/docker/varnish/conf/default.vcl
```diff
@@ -16,7 +16,7 @@ backend default {
 }
 
 # Hosts allowed to send BAN requests
-acl ban {
+acl invalidators {
```
teohhanhui: It seems using the same name as a built-in action throws an error since Varnish 6.0

api/docker/php/docker-entrypoint.sh
```diff
@@ -7,15 +7,18 @@ if [ "${1#-}" != "$1" ]; then
 fi
 
 if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then
-	mkdir -p var/cache var/log var/sessions
+	mkdir -p var/cache var/log
+	setfacl -R -m u:www-data:rwX -m u:"$(whoami)":rwX var
```
teohhanhui: Are we finally ready for `setfacl`? I hope we are. Testers needed for Mac OS and Windows.

dunglas:
I'm afraid to merge this evil PR (#666).

teohhanhui:
I think we should have confirmation from Mac OS and Windows users that the `setfacl` works as expected.

bendavies:
Not in favour of this. Goes against the one process per container docker best practice philosophy

teohhanhui:
@bendavies This builds 3 images, so 3 containers.

teohhanhui:
https://docs.docker.com/compose/compose-file/#target

bendavies:
Apologies I misread the intention!

soyuka:
> I think we should have confirmation from Mac OS and Windows users that the setfacl works as expected.

I'm not using any of these anymore :|.

teohhanhui:
@bendavies No worries! I think that's helpful as it might not be immediately obvious. I've added a note at the top of the `Dockerfile`. :smile:

dunglas:
I can confirm it works on Mac.
Anyone tested with Windows?

maxhelias:
When we build the services it takes the steps from the beginning of the file.
Example with the nginx service, it will build the steps of the service php before and not just the steps that is under the alias

teohhanhui:
@maxhelias That is correct, and it's intended because the nginx image needs to copy files from the php container anyway. :smile:

teohhanhui:
The build steps are cached so there should be no overhead.

maxhelias:
Yes indeed, but the public file is already mounted in volumes in the docker-compose.yml.
And does not that increase the size of the nginx image too much just for this relationship ?

teohhanhui:
We only mount host directories in development. It's not a feasible solution in production.

The earlier stages do not in any way get into the stage being built - the resultant image only contains what's in the target stage.

teohhanhui:
Tested working on Windows. :tada:

soyuka:
Thanks @teohhanhui !

bendavies:
It seems a very annoying consequence of this is that to build the nginx container, the whole php container is build again, just to copy some files. very slow and unnecessary.

PR #678, opened 2018-05-10
Update of react-redux-router dependency

### LINKED ISSUE #675: React Client - Router and Redux-router dependencies mismatch
Hi ! 

I started to work on the client side of my project using this repository configuration. 
I saw in `package.json` that you use 
```
    "react-router-dom": "^4.2.2",
    "react-router-redux": "^4.0.8",
```
But on the github repository of `react-router-redux` they say that you need v5 to work with `react-router`  v4. 
I tried and `const history = syncHistoryWithStore(createBrowserHistory(), store);` is not working on this repository due to this mismatch on dependencies. 

Did you manage to make it work ? 

It's not critical for my project because i don't need it and i changed my package.json dependencies. I only wanted to know if you were aware of this. :)

kayneth:
This fixes #675

Would you prefer to go to `^5.0.0-alpha.9` and explicitly require `"history": "^4.7.2"` then update the code of the index.js to what I've done based on [the documentation
](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux)

Or would you want to go to `react-router` v3 ?

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -5,12 +5,13 @@
   "dependencies": {
     "bootstrap": "^4.0.0",
     "font-awesome": "^4.7.0",
+    "history": "^4.7.2",
     "prop-types": "^15.5.10",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
     "react-redux": "^5.0.6",
     "react-router-dom": "^4.2.2",
-    "react-router-redux": "^4.0.8",
+    "react-router-redux": "^5.0.0-alpha.9",
     "redux": "^3.7.2",
     "redux-form": "^7.0.4",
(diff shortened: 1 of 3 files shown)


dunglas:
I'm not confident requiring alpha versions... Did you try if it works with the generated code?

kayneth:
I tried it and it works. But there's risks using alpha version.. 
Do you prefer to go down to react-router 3 with react-router-redux 4 and `history@^3.3` ?

Furthermore, do we really need to put `react-router-redux` as a dependency ?

dunglas:
If we don’t need it, better drop it.

dunglas:
Closing as we're using connected router now.

PR #703, opened 2018-05-29
Add COC file, remove some files from generated archives

dunglas:
* Add the code of conduct in this repository so GitHub can display it. It's not new one (we have a CoC for years), but I've upgraded to the latest version of Contributor Covenant and replaced the terms "race" and "ethnicity" by "ethnic origin", as discussed here: https://github.com/symfony/symfony-docs/pull/9394#discussion_r172696659 (ContributorCovenant/contributor_covenant#512 / ContributorCovenant/contributor_covenant#476). I'll use the same version in other repositories as well.
* Removed some files from the downloadable archive: LICENSE, the CoC, CONTRIBUTING.md and  README.md as they are pointless when using this repository as a skeleton for a new project.


teohhanhui:
Why are they under `.github` directory? Is there also handling of those files by GitHub? But anyway, I don't think that's a good idea.

soyuka:
> Why are they under .github directory? Is there special handling of those files by GitHub? But anyway, I don't think that's a good idea.

Yes they're then linking these files when you open an issue for example. Anyway according to the docs, these files can also be at the root of the project. I'd be :+1: to left them there (ie not in `.github`) for visibility.

PR #719, opened 2018-06-13
Better markup

dunglas:


README.md
```diff
@@ -1,4 +1,4 @@
-<p align="center"><a href="https://api-platform.com"><img src="https://api-platform.com/logo-250x250.png" alt="API Platform"></a></p>
```
teohhanhui: https://html.spec.whatwg.org/#attr-hx-align

> _**align**_ on **h1**—**h6** elements
>
>     Use CSS instead.

dunglas: Note that align on p is’nt valid either.

teohhanhui: Uhh... 🙈

PR #732, opened 2018-06-26
Improve OPcache settings

ajardin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

`opcache.memory_consumption` is present on the [Symfony documentation](http://symfony.com/doc/current/performance.html#configure-opcache-for-maximum-performance) while `opcache.interned_strings_buffer` should be upgraded also as the default value is quite low.

That's basically exactly the same PR I did [here](https://github.com/dunglas/symfony-docker/pull/24). 😉

### DIFF
--- a/api/docker/php/php.ini
+++ b/api/docker/php/php.ini
@@ -4,6 +4,8 @@ session.auto_start = Off
 short_open_tag = Off
 
 # http://symfony.com/doc/current/performance.html
+opcache.interned_strings_buffer = 16
 opcache.max_accelerated_files = 20000
+opcache.memory_consumption = 256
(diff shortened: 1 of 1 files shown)


teohhanhui:
@ajardin I'm okay with maintaining 2 `php.ini` files for prod and dev separately. The alternative would be to introduce templating, or to use environment variables to set PHP configuration directives.

dunglas:
Thanks @ajardin

PR #751, opened 2018-07-09
Ensure that .env is always loaded in dev

cestcri:
I ran into this because APP_ENV is for whatever reason already set.
This change will lead to less code and increased readability.

### DIFF
--- a/api/public/index.php
+++ b/api/public/index.php
@@ -7,15 +7,16 @@ use Symfony\Component\HttpFoundation\Request;
 
 require __DIR__.'/../vendor/autoload.php';
 
+$env = $_SERVER[‘APP_ENV’] ?? ‘dev’;
+
 // The check is to ensure we don't use .env in production
-if (!isset($_SERVER['APP_ENV'])) {
+if ($env === ‘dev’) {
     if (!class_exists(Dotenv::class)) {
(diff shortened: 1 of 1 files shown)


teohhanhui:
`api/.env` is loaded in `docker-compose.yml`. If you want to use Symfony `DotEnv`, you could just `composer require` it and comment out `APP_ENV` in `api/.env`

cestcri:
Thanks for getting back to me @teohhanhui !
I'm not using Docker, not sure if this is a problem/not supported by the API platform? Maybe I should go for good old Symfony instead of API platform then?

teohhanhui:
If you're not using Docker, then just do `composer req dotenv` :smile:

You shouldn't need to change anything else.

teohhanhui:
Perhaps we should add this to the docs, if it's not already in there.

cestcri:
I did do this (found this on the Symfony doc) but this won't fix the problem if your APP_ENV is already set to dev, which is for whatever reason my case. The suggested change makes this more robust and avoids imho users to struggle for nothing.

teohhanhui:
-1

I understand your problem but I don't think the behaviour should be changed. If you think we could improve the documentation, please open a PR there. Also (more likely) for Symfony documentation.

cestcri:
Fine with me, it was just a suggestion. It's fine on the Symfony doc, that's where I found the hint (which, as described, didn't solve the problem, and which is why I made this suggestion).

PR #759, opened 2018-07-12
Use tmpfs for varnish working directory

teohhanhui:
See https://github.com/docker-library/official-images/pull/4404#issuecomment-401074694


dunglas:
Shouldn't we do the same thing for Kubernetes/Helm @teohhanhui?

teohhanhui:
@dunglas Yes. :)

PR #769, opened 2018-07-17
Delete .env

bendavies:
`.env` is in .gitignore, but not sure how this should work with `docker`, as it's referenced in `docker-compose.yaml`. WDYT @teohhanhui ?


teohhanhui:
I think it shouldn't be in `.gitignore`.

bendavies:
FYI, it's part of the framework bundle flex recipe. https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/manifest.json#L21

And the it ships a `.env.dist` which we do too.

dunglas:
We reference the .env file in `docker-compose.yml`, we cannot delete it just like this. As long as Docker Compose doesn't provide support for dist file, we'll have to remove it from `.gitignore`.

quentinus95:
You can reference the dist file in docker-compose using:
```
env_file:
      - ./.env.dist
```

teohhanhui:
@quentinus95:

> You can reference the dist file in docker-compose

Unfortunately that means it ceases to be a "dist" file. :smile:

teohhanhui:
Closing this one. I think we'll be able to do something about this after https://github.com/symfony/symfony/pull/28533 :smile:

PR #771, opened 2018-07-19
Make the Docker build caching more robust

teohhanhui:
And actually reuse previously built images in CI

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,8 @@ services:
       target: api_platform_php
       cache_from:
         - ${CONTAINER_REGISTRY_BASE}/php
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -8,6 +8,8 @@ services:
       target: api_platform_php
       cache_from:
         - ${CONTAINER_REGISTRY_BASE}/php
+        - ${CONTAINER_REGISTRY_BASE}/nginx
```
teohhanhui: Use the same `cache_from` for the different images built out of the same `Dockerfile`, because:

1. This way we'll reuse the cache regardless of the build order.
2. It has no negative effects. It works.

teohhanhui:
Now the Travis build is faster! :tada:

PR #782, opened 2018-07-27
Add a dev/debug stage using xdebug to the api

### LINKED ISSUE #637: Install Xdebug for development environment
What's the best way for include Xdebug for the development environment and keep clean Dockerfile used for production?

1.  Install Xdebug at the entrypoint.sh
1.  Use a specific Dockerfile.dev (copy of Dockerfile) and install Xdebug (`RUN apk add --no-cache $PHPIZE_DEPS  && pecl install xdebug-2.6.0 && docker-php-ext-enable xdebug`)

I'm using the second solution but need to comment/uncomment line on the `docker-compose.yml` > php service : `dockerfile: Dockerfile.dev`

hschaeidt:
The motivation for this change is to have an easy setup for debugging
using for example PHPStorm as IDE. resolves #637 for example

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,11 +1,12 @@
 # the different stages of this Dockerfile are meant to be built into separate images
 # https://docs.docker.com/compose/compose-file/#target
 
+ARG BUILD_ENV=prod
 ARG PHP_VERSION=7.2
 ARG NGINX_VERSION=1.15
 ARG VARNISH_VERSION=6.0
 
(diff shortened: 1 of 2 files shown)


teohhanhui:
I think this kind of stuff could be easily added in your own project. It's not something I think is important enough to have out of the box.

dunglas:
It definitely deserves a documentation entry however.

hschaeidt:
Thanks, I've opened an issue. So we don't forget about this.

PR #849, opened 2018-10-01
Create React App v2

Fabious:
create-react-app is very soon to be updated to version 2, and adds really cool features, as you can see here : https://github.com/facebook/create-react-app/issues/5103

This PR is meant for api-platform to be ready for it

I used `$ yarn add --exact react-scripts@2.0.1`

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -16,7 +16,7 @@
     "redux-thunk": "^2.2.0"
   },
   "devDependencies": {
-    "react-scripts": "1.1.0"
+    "react-scripts": "2.0.1"
   },
   "scripts": {
     "start": "react-scripts start",
--- a/client/yarn.lock
+++ b/client/yarn.lock
@@ -2,73 +2,1097 @@
 # yarn lockfile v1
 
 
-abab@^1.0.3:
-  version "1.0.4"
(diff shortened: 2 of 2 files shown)


admin/package.json
```diff
@@ -9,12 +9,12 @@
     "react-dom": "^16.3.0"
   },
   "devDependencies": {
-    "react-scripts": "^1.1.0"
+    "react-scripts": "2.0.1"
```
dunglas: I would keep the `^` syntax for both `package.json` files. It will make the maintenance easier on our side.

admin/package.json
```diff
@@ -9,12 +9,18 @@
...
+    "react-scripts": "^2.0.3"
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject"
-  }
+  },
+  "browserslist": [
```
Fabious: Do we keep this ? it's added since create-react-app v2

dunglas: yes please!

dunglas:
Thanks @Fabious! Great job.

PR #856, opened 2018-10-06
Make php-fpm wait for postgres to be ready, even in prod

teohhanhui:


teohhanhui:
Just catching my own mistake. :rofl:

PR #857, opened 2018-10-09
Use shipped php.ini depending on env

Lctrs:
See https://github.com/docker-library/php/issues/692 and https://github.com/docker-library/php/pull/711.

By default, PHP use some unsecure default values when they are not in `php.ini` such as `display_errors = On`.

This PR ensures that we use [`php.ini-production`](https://github.com/php/php-src/blob/master/php.ini-production) in prod and [`php.ini-development`](https://github.com/php/php-src/blob/master/php.ini-development) in dev env to have secure default values in prod.

Here is the diff between dev and prod php.ini : 

[phpini.txt](https://github.com/api-platform/api-platform/files/2459892/phpini.txt)


teohhanhui:
I'd say name them `php.dev.ini` and `php.prod.ini`. WDYT?

teohhanhui:
Uhh, okay, those files are from the Docker official PHP image.

api/Dockerfile
```diff
@@ -52,7 +52,8 @@ RUN set -eux; \
 	apk del .build-deps
 
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
-COPY docker/php/php.ini /usr/local/etc/php/php.ini
+RUN cp $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
+COPY docker/php/api.ini $PHP_INI_DIR/conf.d/
```
teohhanhui: Can we rename it to something like `api-platform.ini`?

api/Dockerfile
```diff
@@ -52,7 +52,8 @@ RUN set -eux; \
 	apk del .build-deps
 
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
-COPY docker/php/php.ini /usr/local/etc/php/php.ini
+RUN cp $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
```
teohhanhui: I think we could just use a symlink?

api/docker/php/docker-entrypoint.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/sh
 set -e
 
+if [ "$APP_ENV" != 'prod' ]; then
+	cp -f ${PHP_INI_DIR}/php.ini-development ${PHP_INI_DIR}/php.ini
```
teohhanhui: I think this has to be done both ways, to present least surprises. Imagine if the user switches between `APP_ENV=dev` and `APP_ENV=prod` on local.

teohhanhui: Also, I think this should be moved into the `if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then` block below.

api/Dockerfile
Lctrs: Why not sticking with the container name ?

api/docker/php/docker-entrypoint.sh
Lctrs: > Also, I think this should be moved into the `if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then` block below.

What if the user runs `docker-compose run php bash` ? It could not end up with the right `php.ini` in this case. But maybe it's not so important in this case. I don't have strong opinion on this.

api/Dockerfile
teohhanhui: Because naming something using the container name inside the container gives no useful information. :laughing:

api/docker/php/docker-entrypoint.sh
teohhanhui: Yeah... But the entrypoint scripts for the Docker official images do not do anything for "other" commands. Indeed, you could never be able to handle everything. So what we care about here are only:

- php-fpm
- php (should be added?)
- bin/console

Lctrs: > 
> 
>
>     * php (should be added?)
> 

Maybe in another PR ?

Lctrs:
@teohhanhui Done. And tests are green. :)

teohhanhui:
Thanks @Lctrs! :tada:

PR #888, opened 2018-11-08
Add support for php cli in docker-entrypoint

Lctrs:
Follow up of #857 .

Will support things like `docker-compose run php php bin/console app:amazing-and-shiny-command`.


dunglas:
Thanks @Lctrs

teohhanhui:
Actually, I think we should remove the handling of both `php` and `bin/console` from the entrypoint script. It's counter intuitive in actual usage.

PR #892, opened 2018-11-13
Minor enhancements for the Docker setup

teohhanhui:


api/docker/php/docker-entrypoint.sh
```diff
@@ -7,19 +7,23 @@ if [ "${1#-}" != "$1" ]; then
...
-		ln -sf ${PHP_INI_DIR}/php.ini-development ${PHP_INI_DIR}/php.ini
 		composer install --prefer-dist --no-progress --no-suggest --no-interaction
-	else
-		ln -sf ${PHP_INI_DIR}/php.ini-production ${PHP_INI_DIR}/php.ini
 	fi
 
-	>&2 echo "Waiting for Postgres to be ready..."
-	until pg_isready --timeout=0 --dbname="${DATABASE_URL}"; do
+	echo "Waiting for db to be ready..."
+	until bin/console doctrine:query:sql "SELECT 1" > /dev/null 2>&1; do
```
dunglas: What's the benefit of this change? It will slower and consume more resources (it can matter in high availability infrastructures)

teohhanhui: Because this uses the Doctrine DBAL, it means that it works with any supported database. Also, it removes the dependency on `postgresql-client`.

> It will slower and consume more resources (it can matter in high availability infrastructures)

Should we do some benchmarks?

dunglas: Bootstrapping Symfony and warming is cache is slow... Btw having pg_client is also nice to debug.

teohhanhui: The cache should be already pre-warmed in `prod` environment so there is no overhead there. As for bootstrapping, I can't say for sure how much difference it makes.

`postgresql-client` is already available in the `db` container. :smile:

dunglas: Yes, but it does not help when you want to debug a connection problem form the PHP app (network issue for instance).

teohhanhui: Hmm... Then one can easily install it, if actually necessary. It definitely should not come with it out of the box.

teohhanhui:
Rebased.

dunglas:
Thanks @teohhanhui

PR #958, opened 2019-01-14
Switch to the new Symfony .env structure

vincentchalamon:


api/public/index.php
```diff
@@ -13,7 +13,7 @@
 }
 
 if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
-    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL);
+    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
```
dunglas: This change should be reverted because it prevents Varnish to work

dunglas:
Thanks @vincentchalamon

bendavies:
what's going on here? we don't use flex >=1.2. it's not even released.

PR #959, opened 2019-01-14
Fix symfony .env

vincentchalamon:
.env* files should not be ignored from docker container


vincentchalamon:
Duplicate from #960

PR #968, opened 2019-01-19
Add optional MongoDB

alanpoulain:
To easily add MongoDB to the distribution.

### DIFF
--- a/api/config/services.yaml
+++ b/api/config/services.yaml
@@ -6,6 +6,14 @@
 parameters:
     env(MERCURE_PUBLISH_URL): ''
     env(MERCURE_JWT_SECRET): ''
+    # Adds fallbacks for MONGODB_URL and MONGODB_DB
+    # if the env vars are not set.
+    # This allows you to run cache:warmup even if your
+    # environment variables are not available yet.
+    # You should not need to change these values.
+    # You can remove them if you are not using MongoDB.
+    env(MONGODB_URL): 'mongodb://api-platform'
+    env(MONGODB_DB): ''
 
 services:
     # default configuration for services in *this* file
--- a/api/docker/php/docker-entrypoint.sh
+++ b/api/docker/php/docker-entrypoint.sh
@@ -21,7 +21,7 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer install --prefer-dist --no-progress --no-suggest --no-interaction
 	fi
 
-	if [ "$MONGODB" != 1 ]; then
+	if [ "$MONGODB_ENABLED" != 1 ]; then
 		echo "Waiting for db to be ready..."
 		until bin/console doctrine:query:sql "SELECT 1" > /dev/null 2>&1; do
 			sleep 1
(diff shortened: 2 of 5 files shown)


api/docker/php/docker-entrypoint.sh
```diff
@@ -21,13 +21,15 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer install --prefer-dist --no-progress --no-suggest --no-interaction
 	fi
 
-	echo "Waiting for db to be ready..."
-	until bin/console doctrine:query:sql "SELECT 1" > /dev/null 2>&1; do
-		sleep 1
-	done
+	if [ "$MONGODB" != 1 ]; then
```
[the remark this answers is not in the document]
alanpoulain: The concept of schema doesn't really exist in MongoDB. This command only updates indexes. I think it can be done manually.

docker-compose.yml
```diff
@@ -68,6 +68,22 @@ services:
     ports:
       - "5432:5432"
 
+#  Uncomment the following lines to use the MongoDB image. Remove them if you don't use MongoDB.
+#  db-mongodb:
+#      # In production, you may want to use a managed database service
+#      image: mongo
```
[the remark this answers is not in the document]
alanpoulain: No I didn't have any auth issue. Maybe you can try to see if you have some?

alanpoulain:
@vincentchalamon The PR for the documentation is here: https://github.com/api-platform/docs/pull/720

api/config/services.yaml
```diff
@@ -6,6 +6,14 @@
 parameters:
     env(MERCURE_PUBLISH_URL): ''
     env(MERCURE_JWT_SECRET): ''
+    # Adds fallbacks for MONGODB_URL and MONGODB_DB
+    # if the env vars are not set.
+    # This allows you to run cache:warmup even if your
+    # environment variables are not available yet.
+    # You should not need to change these values.
+    # You can remove them if you are not using MongoDB.
+    env(MONGODB_URL): 'mongodb://api-platform'
```
vincentchalamon: This variable is never used (in this project configuration), Symfony will throw an exception cause these environment variables are never used

alanpoulain: But if this is removed, you can't require the Doctrine MongoDB ODM bundle without having a cache warmup issue.

vincentchalamon: What is the exception stack trace?

alanpoulain:
Would be better than having to uncomment things and to add build arguments sure.
But in the meantime, I think we could stick to this approach.

dunglas:
It will be a pain to maintain, and will pollute the main repo... This PR is close to be finished, I would prefer to go directly this way.

alanpoulain:
It can be reverted later on, doesn't it? If we wait for the other PR, it means there is no documentation to install / use MongoDB while the new version is already out.

dunglas:
I would prefer a doc entry in the meantime.

api/docker/php/docker-entrypoint.sh
```diff
@@ -21,13 +21,15 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer install --prefer-dist --no-progress --no-suggest --no-interaction
 	fi
 
-	echo "Waiting for db to be ready..."
-	until bin/console doctrine:query:sql "SELECT 1" > /dev/null 2>&1; do
-		sleep 1
-	done
+	if [ "$MONGODB_ENABLED" != 1 ]; then
```
dunglas: You could use both Mongo and Postgres right?

alanpoulain:
Closed in favor of documentation only: https://github.com/api-platform/docs/pull/720

PR #973, opened 2019-01-25
Remove hirak/prestissimo

dunglas:
This package is now useless thanks to Flex.


borNfreee:
could you please point to any resource to get details?

PR #977, opened 2019-01-27
Use again prestissimo

alanpoulain:
This reverts commit cdc399837e05144a32ad0a09c5743edf9f558dbf.
It's because the prefetching of Symfony Flex is only done when the flag `--no-scripts` is not set.
But it needs to be set because the cache clear can't be done (`Could not open input file: bin/console`).


dunglas:
We could install Flex globally instead, as done here: https://github.com/dunglas/symfony-docker/pull/44

alanpoulain:
Closed in favor of: https://github.com/api-platform/api-platform/pull/978

PR #1023, opened 2019-03-02
Bump recipe

maxhelias:
After installing the project, I made an update of the recipes with `composer sync-recipes --force`
That's what it gives, tell me if there are things to remove.

Q | A
-- | --
Bug fix? | no
New feature? | no
BC breaks? | no
Deprecations? | no
Tests pass? | yes
Fixed tickets |  
License | MIT
Doc PR

### DIFF
--- a/api/.env
+++ b/api/.env
@@ -1,7 +1,16 @@
-# This file defines all environment variables that the application needs.
-# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE.
-# Use ".env.local" for local overrides during development.
-# Use real environment variables when deploying to production.
+# In all environments, the following files are loaded if they exist,
+# the later taking precedence over the former:
+#
+#  * .env                contains default values for the environment variables needed by the app
+#  * .env.local          uncommitted file with local overrides
+#  * .env.$APP_ENV       committed environment-specific defaults
+#  * .env.$APP_ENV.local uncommitted environment-specific overrides
+#
+# Real environment variables win over .env files.
+#
+# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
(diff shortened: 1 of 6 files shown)


api/public/index.php
```diff
@@ -13,7 +13,7 @@
 }
 
 if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? $_ENV['TRUSTED_PROXIES'] ?? false) {
-    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL);
+    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_ALL ^ Request::HEADER_X_FORWARDED_HOST);
```
dunglas: This must be reverted. It prevents Varnish to work properly.

maxhelias: It's reverted 😃

maxhelias:
Yes, it's a good idea, I just added it but how to prevent updating the index.php file for Varnish ?
Will not we change the recipe ? Or something else ?

dunglas:
For now will do it manually. Of maybe could we just add a ˋsed` line to remove the extra constant automatically?

maxhelias:
Ok, I understand better now how does the command work.
I delete the file symfony.lock which is used to check the version of the recipe and executed the command to have all the updates (like that of Mercure).
And add the new generated symfony.lock file.

Also the command `composer sync-recipes --force` don't work very well in container php because we don't have the reference Git.

maxhelias:
We don't have the update of index.php file because the symfony.lock file keep the ref and the command check if we have new version of the recipe.
So with the new symfony.lock file, we don't have any probleme with the new update for now

dunglas:
Thank you very much Max!

PR #1041, opened 2019-03-11
Add back missing MERCURE_SUBSCRIBE_URL env var

teohhanhui:
Accidentally removed in #1023.

This also means we need a better smoke test, as the problem was not caught.


teohhanhui:
Looks like we could use [`curl --fail`](https://curl.haxx.se/docs/manpage.html#-f)

PR #1066, opened 2019-03-26
Updated k8s charts

pborreli:


api/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apk add --no-cache \
 		file \
 		gettext \
 		git \
+		postgresql-client \
```
dunglas: to remove

api/docker/nginx/conf.d/default.conf
```diff
@@ -13,6 +13,10 @@ server {
         #set $upstream_host php;
         #fastcgi_pass $upstream_host:9000;
 
+        # Bigger buffer size to handle cache invalidation headers expansion
```
dunglas: To remove, specific to the demo

api/docker/varnish/conf/default.vcl
```diff
@@ -29,11 +33,27 @@ sub vcl_backend_response {
   set beresp.grace = 1h;
 }
 
+sub vcl_recv {
+ if (req.http.X-Blackfire-Query) {
```
dunglas: To remove specific to Blackfire

api/helm/api/templates/secrets.yaml
```diff
@@ -4,15 +4,17 @@ kind: Secret
...
+    app.kubernetes.io/name: {{ include "name" . }}
+    app.kubernetes.io/part-of: {{ include "name" . }}
+    helm.sh/chart: {{ include "chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
 type: Opaque
 data:
   {{ if .Values.postgresql.enabled }}
-  database-url: {{ printf "postgres://%s:%s@%s/%s" .Values.postgresql.postgresUser .Values.postgresql.postgresPassword $postgresqlServiceName .Values.postgresql.postgresDatabase | b64enc | quote }}
+  database-url: {{ printf "pgsql://%s:%s@%s/%s?serverVersion=9.6" .Values.postgresql.postgresqlUsername .Values.postgresql.postgresqlPassword $postgresqlServiceName .Values.postgresql.postgresqlDatabase | b64enc | quote }}
```
dunglas: We should keep `serverVersion` to prevent an extra SQL query at every HTTP request.

pborreli: we didn't remove it, so I guess it's good now

api/helm/api/.helmignore
```diff
@@ -19,3 +19,4 @@
 .project
 .idea/
 *.tmproj
+.vscode/
```
vincentchalamon: There are many other rules to add, you could have a look at a generic global .gitignore configuration: https://gist.github.com/vincentchalamon/d5defad563ed49d9306a4aa57dfd4498

pborreli: yes not mine .. it was inside the demo PR

api/helm/api/values.yaml
```diff
@@ -27,28 +33,51 @@ varnish:
...
+    # kubernetes.io/ingress.class: gce
+    # kubernetes.io/tls-acme: "true"
+  tls:
+    # Secrets must be manually created in the namespace, you can also use cert-manager.
+    # - hosts:
+      # - example.com
+      # - mercure.example.com
+  hosts:
+    api:
+      host: example.com
```
vincentchalamon: `api.example.com` would be more efficient to prevent any conflict with mercure host

pborreli: already discussed with @dunglas but OK to change it if needed

vincentchalamon: What did @dunglas say about it?

pborreli: I think we said that `*.example.com` is different and won't conflict than `example.com`

vincentchalamon: :+1:

soyuka:
thanks @pborreli !

PR #1071, opened 2019-03-27
Use semver for Symfony packages

teohhanhui:


api/composer.json
```diff
@@ -1,28 +1,31 @@
 {
-    "type": "project",
```
teohhanhui: Not needed, as we don't support `composer create-project api-platform/api-platform`

api/composer.json
```diff
@@ -1,28 +1,31 @@
...
-    "license": "proprietary",
```
teohhanhui: Don't assume proprietary license. Let the user take care of this.

teohhanhui:
I know it differs from the Symfony skeleton, but I will not waste my time trying to change things there, as they've already made up their mind (and understandably so, because releases are way more often for Symfony).

But really, as an application developer, semver was revolutionary. We should not be going backwards!

teohhanhui:
> We should keep our skeleton as close a possible from the Symfony one.

Actually, we're nowhere close to Symfony Skeleton: https://github.com/symfony/skeleton/blob/v4.2.4.2/composer.json

teohhanhui:
If I understand correctly, the magic happens here (on `composer create-project`): https://github.com/symfony/flex/blob/v1.2.0/src/Flex.php#L257-L269

Correct me if I'm wrong @nicolas-grekas.

So, since we don't support `composer create-project` in this distribution, it's irrelevant.

nicolas-grekas:
Enterprises want better control of the exact version of their deps. Upgrading to 4.3.0 the day it will be tagged is something that should be explicit, not implicit as in `composer update`.

teohhanhui:
Pretty sure enterprise-minded ones could change their version constraints if that's what they want. And they shouldn't be doing an unqualified `composer update` anyway...

nicolas-grekas:
If you mean removing the `*` for Mercure bundle I agree it should be updated. But removing `4.2.*` looks like a bad idea to me.

teohhanhui:
Requirements:
1. Use semver
2. >Enterprises want better control of the exact version of their deps. Upgrading to 4.3.0 the day it will be tagged is something that should be explicit

Solutions (one of):
1. Don't do unqualified `composer update`
2. Change version constraints to explicitly disallow the next minor version

bendavies:
> > We should keep our skeleton as close a possible from the Symfony one.
> 
> Actually, we're nowhere close to Symfony Skeleton: [symfony/skeleton:composer.json@`v4.2.4.2`](https://github.com/symfony/skeleton/blob/v4.2.4.2/composer.json)

https://github.com/symfony/website-skeleton/blob/v4.2.4.2/composer.json
but you are not close to that either 😆

api/composer.json
vincentchalamon: But `api-platform/api-platform` is a project, not a library. Removing this parameter will set it to the default value `library`. I'd prefer to keep `project` as it is described in composer:
>project: This denotes a project rather than a library. For example application shells like the Symfony standard edition, CMSs like the SilverStripe installer or full fledged applications distributed as packages. This can for example be used by IDEs to provide listings of projects to initialize when creating a new workspace.

vincentchalamon: But `api-platform/api-platform` is a project, not a library. Removing this parameter will set it to the default value `library`. I'd prefer to keep `project` as it is described in composer:
>project: This denotes a project rather than a library. For example application shells like the Symfony standard edition, CMSs like the SilverStripe installer or full fledged applications distributed as packages. This can for example be used by IDEs to provide listings of projects to initialize when creating a new workspace.

api/composer.json
vincentchalamon: `api-platform/api-platform` is supposed to be ready to use, but this value is _highly recommended_ by composer. I'd prefer to keep it filled, even with `proprietary` cause most of users creating a project from this one create a proprietary project (cf. customers).

https://getcomposer.org/doc/04-schema.md#license
>Optional, but it is highly recommended to supply this.

api/composer.json
teohhanhui: > Only use a custom type if you need custom logic during installation. It is recommended to omit this field and have it default to library.

api/composer.json
teohhanhui: `composer validate` does not complain about this. IMO, it's better to not provide a value than the potentially wrong one, since it's optional anyway. And if it's about the license of *this* project, then it should be MIT.

api/composer.json
shehi: > Not needed, as we don't support `composer create-project api-platform/api-platform`
> 
> > Only use a custom type if you need custom logic during installation. It is recommended to omit this field and have it default to `library`.
> 
> https://getcomposer.org/doc/04-schema.md#type

Why not support ```composer create-project ...``` ?!

api/composer.json
```diff
@@ -1,28 +1,31 @@
...
     "require": {
         "php": "^7.1.3",
         "ext-ctype": "*",
         "ext-iconv": "*",
         "api-platform/api-pack": "^1.1",
         "guzzlehttp/guzzle": "^6.3",
-        "symfony/console": "4.2.*",
-        "symfony/dotenv": "4.2.*",
+        "symfony/console": "^4.2",
```
shehi: The line you removed and the one you added aren't the same. Should be ```~4.2.0```. Use tilde instead of caret otherwise all minor versions of v4 will be included (e.g. 4.3, 4.4 etc)

teohhanhui: That's what semver is about. Minor versions are safe to upgrade to, because they're not supposed to introduce breaking changes.

https://semver.org/

shehi: In theory, yes. But I have seen so many breaking packages, even if it something quite minor. Of course that's an unlikely thing to happen to Symfony packages, but regardless, my facts based on my experiences remain. Thanks.

teohhanhui: Yes, I don't deny that. But if you ask me: "Why semver?"

My answer would be, because it encourages consumers (application developers) to update when possible. With minor versions, you wouldn't *expect* BC breaks, but of course we all make mistakes and things get broken. So is it generally safe to update? Yes. Is it absolutely safe to update? Of course not, and it'll never be. But semver reduces friction.

api/composer.json
dunglas: I prefer to keep project too.

api/composer.json
dunglas: Symfony uses MIT: https://github.com/symfony/skeleton/blob/master/composer.json

We should copy this line.

api/composer.json
dunglas: I agree about the rationale (I don’t remember exactly why it’s not done like this in the Symfony skeleton), but we must be in sync with what is in symfony/skeleton (especially to not increase the maintenance burden).

PR #1127, opened 2019-05-03
update-deps.sh: Run composer sync-recipes inside php container

Lctrs:


maxhelias:
The command is not optimal in the container because it does not have the .git folder and Flex ask for each file override

teohhanhui:
@maxhelias Uhh, because the `.git` directory is one level up. Hmm...

teohhanhui:
Anyway, I think yeah, we don't need to change this, as it's only meant to be run by a maintainer.

dunglas:
> Anyway, I think yeah, we don't need to change this, as it's only meant to be run by a maintainer.

Exactly, and it needs manual changes after the command has finished...

PR #1167, opened 2019-06-12
Fix builds

dunglas:
Because of this PR: https://github.com/symfony/recipes/pull/588


dunglas:
See also: https://github.com/symfony/symfony/issues/31497#issuecomment-501419860

bendavies:
be careful. I don't think this works
https://github.com/symfony/flex/issues/476

PR #1173, opened 2019-06-14
Dotenv instance is constructed with unnecessary argument

bigwhoop:
\Symfony\Component\Dotenv\Dotenv has no constructor that could take arguments


antograssiot:
It does in 4.3 
https://github.com/symfony/dotenv/blob/4.3/Dotenv.php#L45

bigwhoop:
hi! yup, looks like the 2.4.4 release from 2 days ago fixed it by upgrading to symfony/dotenv 4.3. I checked against 2.4.3 that installed symfony/dotenv 4.2.8 where no constructor was added yet. my bad, sorry for the confusion.

PR #1182, opened 2019-06-27
Use https link to symfony doc

Lctrs:


teohhanhui:
This is pointless as symfony.com sets the `Strict-Transport-Security` header which causes a redirect to `https://` anyway.

teohhanhui:
But sure, it's harmless.

PR #1183, opened 2019-06-28
Disable OPCache timestamps check in prod

Lctrs:


dunglas:
Good idea, this PR probably needs to be updated now that we have a prod setup. /cc @teohhanhui

Lctrs:
@dunglas seems like a rebasing is enough. The prod setup hasn't modified this part.

soyuka:
feel free to merge @teohhanhui

teohhanhui:
Something is definitely wrong here.

teohhanhui:
Thanks @Lctrs! :tada:

PR #1236, opened 2019-08-29
Add debian php image version

d0niek:
Our app has to connect to HANA via odbc but our driver doesn't work with alpine so I had to use debian version.
Now we can switch distro by filling **arg** `ALIPNE` with `-alipne` or leave it empty.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -8,27 +8,58 @@ ARG PHP_VERSION=7.3
 ARG NGINX_VERSION=1.17
 ARG VARNISH_VERSION=6.2
 
+# Leave it blank to use debian
+# fill with "-alpine" to use alpine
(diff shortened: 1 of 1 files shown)


dunglas:
I fear that it adds complexity and increase the maintenance burden, wouldn't it better to explain this in the documentation?

d0niek:
I think one maintenance problem is that when you install new `PERSISTENT_DEPS` or new lib for php extensions and there are different names for alpine and debian.

soyuka:
> I fear that it adds complexity and increase the maintenance burden, wouldn't it better to explain this in the documentation?

:+1: at least it should be a new Dockerfile I don't think that we can maintain 2 oses at the same time

teohhanhui:
This is the wrong approach.

PR #1237, opened 2019-09-01
Remove quotes from TRUSTED_HOSTS

jamesisaac:
Maybe I'm missing something, but this just ends up with single quotes in the regex, so it won't match any hostnames (on Windows at least).

See [CORS_ALLOW_ORIGIN](https://github.com/api-platform/api-platform/blob/767e635967019577952df7592725969c37b108f4/api/.env#L34) in the same file for an example of regex, and no other values in the `.env` are wrapped in single quotes.

### DIFF
--- a/api/.env
+++ b/api/.env
@@ -20,7 +20,7 @@ VARNISH_URL=http://cache-proxy
 APP_ENV=dev
 APP_SECRET=!ChangeMe!
 TRUSTED_PROXIES=10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
-TRUSTED_HOSTS='^localhost|api$'
+TRUSTED_HOSTS=^localhost|api$
 ###< symfony/framework-bundle ###
 
 ###> doctrine/doctrine-bundle ###


teohhanhui:
Yes, you're absolutely right. This was a mistake due to the special format of the Docker / Docker Compose `.env` file.

teohhanhui:
Thanks @jamesisaac! :tada:

jamesisaac:
Great!  Thanks for the speedy follow up

jamesisaac:
@teohhanhui @dunglas I think this issue was brought back again in https://github.com/api-platform/api-platform/commit/56c683cc855fb44f9f5e5db92a123394959cd0e9#diff-18ebc550c4e057971c5a2335cdf1a3eb

teohhanhui:
Hmm... Actually, looking back, it seems perfectly fine to have quotes here, as long as this file is not loaded by Docker Compose, which does not follow the same rules.

jamesisaac:
Ahh I see, that may have been what happened.  There have been projects where we didn't need the "monorepo" structure so moved the `/api` folder up into the root, so this `.env` file might have started being read by Docker Compose (taking precedence over Symfony), and I think this is the only line which is incompatible with DC's format.

Not 100% sure that's what happened but I think that's most likely explanation.  Sorry for the confusion!

PR #1276, opened 2019-10-05
Changed url to use secure links if supported

wenzdey:
HAL stateless.co domain is suspended. Changed HAL url to their GitHub.


alanpoulain:
Thanks @wenzdey.

PR #1286, opened 2019-10-10
Switch to the official varnish Docker image

teohhanhui:
Since there's an [official `varnish` Docker image](https://hub.docker.com/_/varnish) now, we should use it and abandon maintaining [our own image](https://hub.docker.com/r/cooptilleuls/varnish).

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -6,7 +6,7 @@
 # https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG PHP_VERSION=7.3
 ARG NGINX_VERSION=1.17
-ARG VARNISH_VERSION=6.2
+ARG VARNISH_VERSION=6.3
 
 
 # "php" stage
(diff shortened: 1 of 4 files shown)


api/Dockerfile
```diff
@@ -119,6 +119,8 @@ COPY --from=api_platform_php /srv/api/public ./
 
 # "varnish" stage
 # does not depend on any of the above stages, but placed here to keep everything in one Dockerfile
-FROM cooptilleuls/varnish:${VARNISH_VERSION}-alpine AS api_platform_varnish
+FROM varnish:${VARNISH_VERSION} AS api_platform_varnish
 
-COPY docker/varnish/conf/default.vcl /usr/local/etc/varnish/default.vcl
+COPY docker/varnish/conf/default.vcl /etc/varnish/default.vcl
+
+CMD ["varnishd", "-F", "-f", "/etc/varnish/default.vcl", "-p", "http_resp_hdr_len=65536", "-p", "http_resp_size=98304"]
```
teohhanhui: Copied from the Helm chart.

soyuka: could we add a comment for that?

PR #1289, opened 2019-10-12
Wait 180 seconds until database is ready or die.

Niko9911:
Wait 30 seconds until database is ready otherwise exit with error code of 1.

### DIFF
--- a/api/docker/php/docker-entrypoint.sh
+++ b/api/docker/php/docker-entrypoint.sh
@@ -22,8 +22,16 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 	fi
 
 	echo "Waiting for db to be ready..."
+	REPEAT=0
 	until bin/console doctrine:query:sql "SELECT 1" > /dev/null 2>&1; do
+		echo -ne '.'
+		REPEAT=$((REPEAT + 1))
 		sleep 1
+
+		if [ $REPEAT -eq 15 ]; then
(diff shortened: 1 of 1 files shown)


dunglas:
It's 15 seconds isn't it?

Niko9911:
Yes, and according to my testing It needs to be raised to 90 seconds at least, so 120 to be sure. I will make a fix.

Niko9911:
Just to be on safe side, I did put 180 seconds. 

This timeout idea is anyways getting our deployments stuck. Instead of getting stuck due a configuration mistake, it will exit unclean.

teohhanhui:
This is not the correct approach. Please use the healthcheck feature of your container orchestrator.

Niko9911:
@teohhanhui Please show me how you do it in docker-compose and add it into the repo. 

Youre now expecting everyone to run this in kubernetes or what?

That is totally not understandable point of view.

teohhanhui:
But we could add a default `HEALTHCHECK` instruction to our `php` image indeed. It'd require `cgi-fcgi` (available from https://pkgs.alpinelinux.org/package/v3.10/main/x86_64/fcgi).

teohhanhui:
> Youre now expecting everyone to run this in kubernetes or what?

No. You can use healthcheck in Docker Compose too. :smile:

> Please show me how you do it in docker-compose and add it into the repo.

Yeah, we should do that.

teohhanhui:
You can find an example here: https://github.com/renatomefi/php-fpm-healthcheck

PR #1291, opened 2019-10-14
Feature/issue 1196 imagepullsecrets

### LINKED ISSUE #1196: Helm chart imagePullSecrets
It's been updated in the standard helm templates to include the variable `imagePullSecrets` - the variable would be a "nice to have" feature, so that in the helm chart, if the image is pushed to a private repo, a developer can configure a secret to use to authorise the image pull.

See:
https://github.com/helm/helm/pull/5271/files

Otherwise we end up modifying each chart or using the feature so that the pull secret is added to every pod in the namespace as described here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

Mathieudewet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #1196 
| License       | MIT
| Doc PR        |

### DIFF
--- a/api/helm/api/values.yaml
+++ b/api/helm/api/values.yaml
@@ -90,3 +90,5 @@ resources: {}
   # requests:
   #  cpu: 100m
   #  memory: 128Mi
+imagePullSecrets:
+  # See https://github:com/helm/helm/pull/5271/files This is for work with private repositories.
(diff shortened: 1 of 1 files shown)


dunglas:
Have you tested that it works? In https://github.com/helm/charts/search?q=imagePullSecret&unscoped_q=imagePullSecret it looks like this value is just commented. Also I don't think that a link to the Helm's PR is useful.

PR #1298, opened 2019-10-17
NPM: Use exact version

mvegter:
This should prevent any unexpected changes to dependencies, which can affect the ability reproduce builds.

Instead of using latest, you could use something like Renovate. Renovate Automatically updates npm, yarn, meteor, docker and meteor dependencies via Pull Requests. And is free for open-source projects.


dunglas:
See https://github.com/api-platform/website/pull/183

PR #1321, opened 2019-10-31
Add healthcheck php-fpm

### LINKED ISSUE #1295: Include healthcheck for php container out of the box
**Description**  
Currently, we don't include a healthcheck for the `php` container out of the box. This could be a pain when the container doesn't start for any reason, and the container orchestrator doesn't know when to kill it (in the entrypoint script, we wait for the database to become available), so the service would remain unavailable.

See discussion from https://github.com/api-platform/api-platform/pull/1289#issuecomment-542633436 onwards.

**Example**  
N/A

patrick-mota:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #1295  
| License       | MIT
| Doc PR        | 

Add healthcheck in php container see #1295

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -15,6 +15,7 @@ FROM php:${PHP_VERSION}-fpm-alpine AS api_platform_php
 # persistent / runtime deps
 RUN apk add --no-cache \
 		acl \
+		fcgi \
 		file \
 		gettext \
 		git \
@@ -56,8 +57,15 @@ RUN set -eux; \
 	apk del .build-deps
 
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
+
 RUN ln -s $PHP_INI_DIR/php.ini-production $PHP_INI_DIR/php.ini
 COPY docker/php/conf.d/api-platform.prod.ini $PHP_INI_DIR/conf.d/api-platform.ini
+
+RUN set -eux; \
+	{ \
+		echo '[www]'; \
+		echo 'ping.path = /ping'; \
+	} | tee /usr/local/etc/php-fpm.d/docker-healthcheck.conf
 
 # Workaround to allow using PHPUnit 8 with Symfony 4.3
 ENV SYMFONY_PHPUNIT_VERSION=8.3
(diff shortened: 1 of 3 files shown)


dunglas:
Could you also add a similar `livenessProbe` and a `readynessProbe` in (https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) in the Helm chart? https://github.com/api-platform/api-platform/blob/master/api/helm/api/templates/php-deployment.yaml (/cc @pborreli)

api/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apk add --no-cache \
 		file \
 		gettext \
 		git \
+		fcgi \
```
teohhanhui: Please sort alphabetically.

api/Dockerfile
```diff
@@ -54,6 +55,13 @@ RUN set -eux; \
 	apk add --no-cache --virtual .api-phpexts-rundeps $runDeps; \
 	\
 	apk del .build-deps
+	
+HEALTHCHECK --interval=10s --timeout=3s \
+   	CMD \
```
teohhanhui: Please be mindful of the whitespace in this file. Use tabs, not spaces, for indentation.

api/Dockerfile
```diff
@@ -54,6 +55,13 @@ RUN set -eux; \
...
+    SCRIPT_NAME=/ping \
```
teohhanhui: Are you sure this works? We need to set `ping.path` in the php-fpm config: https://www.php.net/manual/en/install.fpm.configuration.php#ping.path

> If this value is not set, no URI will be recognized as a ping page.

api/Dockerfile
teohhanhui: I'd suggest 1 tab for `CMD`, and 2 tabs for each line of the command. WDYT?

teohhanhui: Looking at the official healthcheck examples: https://github.com/docker-library/healthcheck

We should use a separate script file instead.

teohhanhui:
DX impact: the container's log would be filled with lines like:

```
127.0.0.1 -  19/Nov/2019:17:03:27 +0000 "GET /ping" 200
```

teohhanhui:
Thanks @patrick-mota! :tada:

PR #1335, opened 2019-11-19
Replace h2-proxy by Vulcain

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

Allows to use [the Vulcain protocol](https://vulcain.rocks) with API Platform and simplify the Docker configuration by getting rid of an extra layer (h2-proxy) for the JS clients and Mercure.


Dockerfile.tls
```diff
@@ -1,17 +1,17 @@
+# Use this self-generated certificate only in dev, IT IS NOT SECURE!
```
teohhanhui: Could we move + rename this file? Perhaps `dev-tls/Dockerfile`?

Dockerfile.tls
```diff
@@ -1,17 +1,17 @@
...
+
+
 FROM alpine:latest
 
 RUN apk add --no-cache openssl
+WORKDIR "/cert"
 
-# Use this self-generated certificate only in dev, IT IS NOT SECURE!
 RUN openssl genrsa -des3 -passout pass:NotSecure -out cert.pass.key 2048
```
teohhanhui: The openssl `genrsa` command is deprecated. We should use the `genpkey` command, e.g. https://github.com/api-platform/demo/blob/55f967d80b811555840fd20e20a9f2a3bab999d1/api/docker/php/docker-entrypoint.sh#L19

api/.env
```diff
@@ -40,5 +40,5 @@ TRUSTED_HOSTS='^localhost|api$'
 ###< symfony/framework-bundle ###
 
 # API Platform distribution
-MERCURE_SUBSCRIBE_URL=http://localhost:1337/hub
+MERCURE_SUBSCRIBE_URL=https://localhost:1337/.well-known/mercure
```
teohhanhui: Hmm... Isn't the point of `.well-known` URLs so that it does not need to be specified anymore? So we should only pass the host, no?

teohhanhui:
It kinda sucks for DX that `http://` URLs stop working... So the user can't just enter `localhost` in the address bar. :disappointed:

dunglas:
I agree but on the other hand it’s at the same time good for DX because it simplifies everything.

Maybe can we add a command in the getting started guide to add this certificate to the CA bundle or something like that for DX?

api/.env
dunglas: Using the well known URL is just a SHOULD in the spec. So I think we have to keep this option.

teohhanhui:
> at the same time good for DX because it simplifies everything.

I disagree. It removes a dev HTTP proxy, which simplifies maintenance. But DX is worse.

teohhanhui:
I think we **should** keep both HTTP / HTTPS working.

dunglas:
What's the benefit? The admin and the client are **currently** not working without an HTTPS API anyway. CRA doesn't support bot HTTP and HTTPS at the same time, so we need an extra layer, which is complex to maintain, and increase the perceived complexity of the Docker configuration for the end user too.

docker/dev-tls/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# persistent / runtime deps
+RUN apk add --no-cache \
+		nss-tools \
+	;
+
+WORKDIR /certs
+
+ARG MKCERT_VERSION=1.4.1
+RUN set -eux; \
+	wget -O /usr/local/bin/mkcert https://github.com/FiloSottile/mkcert/releases/download/v$MKCERT_VERSION/mkcert-v$MKCERT_VERSION-linux-amd64; \
```
dunglas: @teohhanhui what's the benefit of using mkcert instead of the native `openssl` package? It's an extra external thing to maintain.

Can't we at least find a trick to use mkcert as intended: to add the CA to the local trust store?

teohhanhui: `mkcert` has two use cases: certificate generation, and installing a root CA certificate. We're doing the former. The benefit over running a bunch of `openssl` commands is that it's guaranteed to be doing The Right Thing(TM), so it's one **less** thing for us to maintain. It's recommended a lot in the CRA issues I've seen.

teohhanhui: > use mkcert as intended: to add the CA to the local trust store?

This always has to be done explicitly by the user, for security reasons as you've explained before.

teohhanhui:
> What's the benefit?

> the user can't just enter localhost in the address bar. :disappointed:

This will bite a lot of users. I'm sure we'll get a lot of support questions because of this... (even though the user might figure it out a few seconds later...)

teohhanhui:
> so we need an extra layer, which is complex to maintain

There is a (simple?) alternative, if we use https://github.com/jwilder/nginx-proxy we just need to set the correct env vars on the containers. It's automated.

api/.env
teohhanhui: Hmm... But in the MercureBundle, it should default to the well known URL, such that it's sufficient to specify only the host.

dunglas: But why if you expose the hub at `https://hub.example.com/` as allowed by the spec? The bundle will transform it to `https://hub.example.com/.well-known/mercure` and it will not work.

docker/dev-tls/Dockerfile
dunglas: Ok let's keep `mkcert`. 

> This should always be done explicitly by the user

I agree, but then we can use a volume trick or something like that to check if the user has done this, and ask the user to do this in the docs.

dunglas:
Thanks for helping @teohhanhui!

api/.env
teohhanhui: Okay, I need to read the spec... :laughing:

docker/dev-tls/Dockerfile
teohhanhui: Yeah, I was thinking we should generate the root CA cert in an easily accessible location.

PR #1337, opened 2019-11-20
Skip vulcain headers from varnish request

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no ^^ 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/api/docker/varnish/conf/default.vcl
+++ b/api/docker/varnish/conf/default.vcl
@@ -32,6 +32,9 @@ sub vcl_recv {
 
   # Remove the "Forwarded" HTTP header if exists (security)
   unset req.http.forwarded;
+  # Remove "Preload" and "Fields" HTTP header to improve Vulcain's performance
+  unset req.http.preload;
+  unset req.http.fields;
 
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks @soyuka

api/docker/varnish/conf/default.vcl
```diff
@@ -32,6 +32,9 @@ sub vcl_recv {
 
   # Remove the "Forwarded" HTTP header if exists (security)
   unset req.http.forwarded;
+  # Remove "Preload" and "Fields" HTTP header to improve Vulcain's performance
```
teohhanhui: Hmm... Actually, Varnish should be able to tell that it's safe to not vary the cache based on these headers if there's no `Vary: Preload` and `Vary: Fields` in the response from the backend (php).

teohhanhui: /cc @dunglas

PR #1422, opened 2020-02-18
Keep LICENSE in archive

### LINKED ISSUE #1413: LICENSE is missing in release package
This is more kind of a **HINT**

**Description**  
When downloading the release package, the LICENSE file is missing.

**Example**  
Download the https://github.com/api-platform/api-platform/archive/v2.5.4.tar.gz and unpack it. The folder will not contain the License.

Bildschirmfoto von 2020-02-11 14-07-13

**Why**
I think it should be there because the License states

> The above copyright notice and this permission notice shall be included in all
copies

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #1413 
| License       | MIT
| Doc PR        | na

I guess that this line skips the License in the archive.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -18,6 +18,5 @@ composer.lock text eol=lf merge=ours
 
 .github export-ignore
 .travis.yml export-ignore
-LICENSE export-ignore
 README.md export-ignore
 update-deps.sh export-ignore


teohhanhui:
Thanks @soyuka! :scroll:

PR #1433, opened 2020-02-26
Update Create React App

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fino
| License       | MIT
| Doc PR        | https://github.com/api-platform/docs/pull/1054


admin/public/manifest.json
```diff
@@ -1,14 +1,24 @@
 {
-  "short_name": "API Platform Admin",
-  "name": "Admin",
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
alanpoulain: To revert?

dunglas: Good catch

PR #1434, opened 2020-02-26
CORS: whitelist Vulcain headers

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/api/config/packages/nelmio_cors.yaml
+++ b/api/config/packages/nelmio_cors.yaml
@@ -3,7 +3,7 @@ nelmio_cors:
         origin_regex: true
         allow_origin: ['%env(CORS_ALLOW_ORIGIN)%']
         allow_methods: ['GET', 'OPTIONS', 'POST', 'PUT', 'PATCH', 'DELETE']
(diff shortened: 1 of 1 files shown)


api/config/packages/nelmio_cors.yaml
```diff
@@ -3,7 +3,7 @@ nelmio_cors:
         origin_regex: true
         allow_origin: ['%env(CORS_ALLOW_ORIGIN)%']
         allow_methods: ['GET', 'OPTIONS', 'POST', 'PUT', 'PATCH', 'DELETE']
-        allow_headers: ['Content-Type', 'Authorization']
+        allow_headers: ['Content-Type', 'Authorization', 'Preload', 'Fields']
```
maxhelias: Maybe we should update the recipe too ?

dunglas: Maybe, but I'm not sure that Symfony wants to support Vulcain in the recipes right now.

PR #1447, opened 2020-03-05
[Configuration] update default.conf for add a port 80

DurandSacha:

### DIFF
--- a/api/docker/nginx/conf.d/default.conf
+++ b/api/docker/nginx/conf.d/default.conf
@@ -1,4 +1,5 @@
 server {
+    listen      80;
     root /srv/api/public;
 
     location / {


api/docker/nginx/conf.d/default.conf
```diff
@@ -1,4 +1,5 @@
 server {
+    listen      80;
```
soyuka: not sure we want this, what was your issue?

DurandSacha: There is no issue referenced in this regard. It doesn't bring any additional functionality, but I think it's a useful clarification

soyuka:
Ports are handled by docker, we'll leave the default value in this configuration file. Feel free to change it on your project if needed.

PR #1451, opened 2020-03-10
Enable inline cache for buildx bake

teohhanhui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A


teohhanhui:
But we probably want to switch to using cache manifest now that quay.io supports [Image Manifest v2, Schema v2](https://docs.docker.com/registry/spec/manifest-v2-2/): https://groups.google.com/d/msg/quay-sig/lImWndQIrE0/2mRVOJOhBQAJ

PR #1478, opened 2020-04-01
Improve Helm chart and upgrade to v3

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Upgrade the Helm chart for v3. Also improve its general architecture.

### DIFF
--- a/api/helm/api/templates/php-deployment.yaml
+++ b/api/helm/api/templates/php-deployment.yaml
@@ -1,9 +1,10 @@
+{{- $fullName := include "api.fullname" . -}}
 {{- $name := "php" -}}
 {{- $data := dict "name" $name "Chart" .Chart "Release" .Release "Values" .Values -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ include "api.fullname" . }}-{{ $name }}
+  name: $fullName-{{ $name }}
   labels:
     {{- include "api.labels" $data | nindent 4 }}
 spec:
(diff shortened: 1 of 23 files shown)


dunglas:
I used the default chart generated by `helm create` as base.

api/helm/api/templates/php-deployment.yaml
```diff
@@ -1,89 +1,102 @@
...
                   key: mercure-subscribe-url
             - name: MERCURE_JWT_TOKEN
               valueFrom:
                 secretKeyRef:
-                  name: {{ template "fullname" . }}
+                  name: {{ include "api.fullname" . }}
                   key: mercure-jwt-token
+          ports:
+            - containerPort: 9000
+          # TODO: add livenessProbe and readinessProbe using the cgi-fcgi command
```
jewome62, line 88: using the cgi-fcgi command ? you can just check tcp port is open ?
```
          readinessProbe:
            tcpSocket:
              port: 9000
            initialDelaySeconds: 5
            periodSeconds: 3
          livenessProbe:
            tcpSocket:
              port: 9000
            initialDelaySeconds: 5
            periodSeconds: 3
```

api/helm/api/values.yaml
```diff
@@ -18,26 +19,38 @@ php:
...
   enabled: true
-  #url: https://example.com
-  repository: quay.io/api-platform/varnish
-  tag: latest
-  pullPolicy: Always
+  url: https://example.com
+  image:
+    repository: quay.io/api-platform/varnish
+    tag: latest
+    pullPolicy: IfNotPresent
```
jewome62, line 42: you can't use IfNotPresent with latest. because it will never update image.
But that's part override by value into command helm update

api/helm/api/templates/php-deployment.yaml
dunglas: I don't know if it's good enough. I'll try. WDYT @teohhanhui?

dunglas: Added, it looks like it works properly! Thanks

PR #1492, opened 2020-04-10
Fix SSL_ERROR_RX_RECORD_TOO_LONG

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT

Not sure it's the right fix...

I have an `SSL_ERROR_RX_RECORD_TOO_LONG` error while using following configuration:
```yaml
# docker-compose.yml
services:
  client:
    build:
      target: api_platform_client_nginx
    ports:
      - target: 80
        published: 443
        protocol: tcp
```

But it's working with this PR fixes and the following configuration:
```yaml
# docker-compose.yml
services:
  client:
    build:
      target: api_platform_client_nginx
    ports:
      - target: 443
        published: 443
        protocol: tcp
```

### DIFF
--- a/admin/docker/nginx/conf.d/default.conf
+++ b/admin/docker/nginx/conf.d/default.conf
@@ -1,5 +1,10 @@
 server {
+    listen 443 ssl;
+
     root /usr/src/admin/build;
+
+    ssl_certificate /usr/src/client/node_modules/webpack-dev-server/ssl/localhost.crt;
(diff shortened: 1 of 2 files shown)


vincentchalamon:
The better approach is to use nginx-proxy like in [api-platform/docker-compose-prod](https://github.com/api-platform/docker-compose-prod)

PR #1500, opened 2020-04-15
Fixing CI disabling entrypoint

pborreli:
Will probably fail just after (can't reproduce locally, maybe a permission issue with alpine or runner)


soyuka:
Thanks @pborreli

PR #1520, opened 2020-05-02
Helm, use replica count of PHP for php-deployment

blazarecki:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT
Update the php deployment to use the php replica count value instead of the Nginx one.


dunglas:
Oops. Merci Benjamin!

PR #1553, opened 2020-06-07
fix helm php-deployment.yaml initContainers

florinceleapca:
There is no /bin/bash in the image, use /bin/sh

| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | yes/no
| Tickets       | fixes #... 
| License       | MIT


dunglas:
Good catch, thanks!

PR #1574, opened 2020-07-01
Add opcache preload config for Production.

gries:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

This adds opcache preload config for the prod php.ini as described [here](https://symfony.com/doc/current/performance.html#use-the-opcache-class-preloading)
It would be great to see this in 2.5.7 but I have created the PR against master because technically it's a new feature I guess.


dunglas:
Awesome, thanks! It will be in the next patch release (we only have a master branch for the distribution).

gries:
Thanks for merging and for api-platform in general 😅

PR #1614, opened 2020-08-08
Fix volume mapping with Docker host

### LINKED ISSUE #1155: Docker: how to make node_modules visible?
Breaking my head on this for a while now.

When I download release 2.4.3 and run the docker instance, the `/client/node_modules` and `/admin/node_modules folder` on the host are empty folders (owned by root). In the containers, both folders are populated with their packages.
```
    volumes:
      - ./admin:/usr/src/admin:rw,cached
      - /usr/src/admin/node_modules
```

The issue i'm having is that this the IDE cannot see the packages, which obviously gives me all kinds of errors.

What i would like to achieve is to have the node_modules folder from the container visibile in the host, so when i add a package in the container, this is visible in the host folder as well. I do not need to make changes in the host node_modules folder.

I tried changing the above lines to the below:
```
    volumes:
      - ./admin:/usr/src/admin:rw,cached
      - ./admin/node_modules:/usr/src/admin/node_modules
```
But this results in `/bin/sh: react-scripts: not found`

I also tried the solution in [this](https://stackoverflow.com/questions/51097652/install-node-modules-inside-docker-container-and-synchronize-them-with-host) link, but this only results on errors on my side.

What should be the preferred way to handle this? Maybe I could add a PR for the docs if there is one.

grimzy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #1155
| License       | MIT

Implemented the changes mentioned in https://stackoverflow.com/a/52092711/2657607 to fix #1155. 

Tested on macOS 10.14.6 with Docker Desktop 2.3.0.4.

`node_modules` directories are now visible in the IDE:

- I don't know how to add tests for the docker build
- I don't think this breaks backward compatibility but this PR removes the `VOLUME` from the "development" stages of the node Dockerfiles

### DIFF
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -10,7 +10,9 @@ ARG NGINX_VERSION=1.17
 # "development" stage
 FROM node:${NODE_VERSION}-alpine AS api_platform_admin_development
 
-WORKDIR /usr/src/admin
+# create and define the node_modules's cache directory
+RUN mkdir -p /usr/src/cache
+WORKDIR /usr/src/cache
 
 # prevent the reinstallation of node modules at every changes in the source code
 COPY package.json yarn.lock ./
@@ -23,13 +25,16 @@ RUN set -eux; \
 	yarn install; \
 	apk del .gyp
 
-COPY . ./
+WORKDIR /usr/src/admin
 
-VOLUME /usr/src/admin/node_modules
+COPY . ./
 
 ENV HTTPS true
 
-CMD ["yarn", "start"]
+COPY docker/node/start-cmd.sh /usr/local/bin/start-cmd
+RUN chmod +x /usr/local/bin/start-cmd
+
+CMD ["start-cmd"]
 
 
 # "build" stage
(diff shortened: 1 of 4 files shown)


dunglas:
Hasn't this change a bad performance impact? Initially we decided to not expose the `node_modules` directory because exposing it through a volume was slow as hell on Mac.

grimzy:
It does; both when starting the container and at runtime. Although runtime performance is not too bad for development (only developed with this setup for a week).

Should I instead document this somewhere in the [docs](https://github.com/api-platform/docs)?

dunglas:
I think a comment and an entry in the docs would be better indeed.

romano1996:
Is https://github.com/api-platform/api-platform/issues/1686 related?

dunglas:
We totally refactored this part. Closing for now, feel free to reopen if still needed on the new setup.

PR #1629, opened 2020-08-23
add more logs to docker php enttrypoint

MaximeThoonsen:
| Q             | A
| ------------- | ---
| Bug fix?      | no 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

We had some difficulties debuging an app. The error was that the php container didn't have the rights to reach the database in AWS (wrong security group documentation). 
The logs didn't help us enought to find the issue quickly. I suggest adding these two lines to help other people that may have the same problem to find out what's going on more easily.

### DIFF
--- a/api/docker/php/docker-entrypoint.sh
+++ b/api/docker/php/docker-entrypoint.sh
@@ -27,9 +27,19 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 	fi
 
 	echo "Waiting for db to be ready..."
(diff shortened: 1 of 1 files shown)


dunglas (approved):
I'm not very happy with this loop. This should timeout at some point if the DBMS isn't available.

Maybe could we do something like that: https://stackoverflow.com/a/12321815

MaximeThoonsen:
Yes you're right. I've updated the PR. I tried to write something very easy to understand.

dunglas:
LGTM! Thank you.

PR #1630, opened 2020-08-24
Fix helm configuration

vincentchalamon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tickets       |
| License       | MIT
| Doc PR        |

Helm configuration tested and fixed at https://github.com/api-platform/demo/pull/155

### DIFF
--- a/api/helm/api/Chart.yaml
+++ b/api/helm/api/Chart.yaml
@@ -7,8 +7,8 @@ home: https://api-platform.com
 icon: https://api-platform.com/logo-250x250.png
 dependencies:
     - name: postgresql
-      version: ~8.6.0
-      repository: https://charts.bitnami.com/bitnami
+      version: ~9.3.2
+      repository: https://charts.bitnami.com/bitnami/
       condition: postgresql.enabled
     - name: mercure
       version: ~3.0.0
--- a/api/helm/api/templates/configmap.yaml
+++ b/api/helm/api/templates/configmap.yaml
@@ -1,3 +1,4 @@
+{{- $fullName := include "api.fullname" . -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -8,8 +9,8 @@ data:
   env: {{ .Values.php.env | quote }}
   debug: {{ .Values.php.debug | quote }}
   cors-allow-origin: {{ .Values.php.corsAllowOrigin | quote }}
-  varnish-url: {{ if .Values.varnish.url }}{{ .Values.varnish.url | quote }}{{ else }}http://varnish{{ end }}
+  varnish-url: {{ if .Values.varnish.url }}{{ .Values.varnish.url | quote }}{{ else }}{{ printf "http://%s-varnish" $fullName }}{{ end }}
   trusted-hosts: {{ .Values.php.trustedHosts | quote }}
   trusted-proxies: {{ join "," .Values.php.trustedProxies }}
-  mercure-publish-url: {{ .Values.mercure.publishUrl | quote }}
-  mercure-subscribe-url: {{ .Values.mercure.subscribeUrl | quote }}
+  mercure-publish-url: {{ printf "http://%s-mercure/.well-known/mercure" .Release.Name | quote }}
+  mercure-subscribe-url: {{ printf "https://%s/.well-known/mercure" (index .Values.ingress.hosts 1).host | quote }}
--- a/api/helm/api/templates/ingress.yaml
+++ b/api/helm/api/templates/ingress.yaml
@@ -27,19 +27,23 @@ spec:
 {{- end }}
   rules:
   {{- range .Values.ingress.hosts }}
-    - host: {{ .host | quote }}
+    {{ $host := .host }}
+    - host: {{ $host | quote }}
       http:
         paths:
         {{- range .paths }}
           - path: {{ . }}
             backend:
-              {{- if $.Values.varnish.enabled }}
-              serviceName: {{ $fullName }}-varnish
-              servicePort: {{ $.Values.varnish.service.port }}
+              {{- if contains "mercure" $host }}
+                serviceName: {{ printf "%s-mercure" $.Release.Name }}
+                servicePort: {{ $.Values.mercure.service.port }}
+              {{- else }}{{- if $.Values.varnish.enabled }}
+                serviceName: {{ $fullName }}-varnish
+                servicePort: {{ $.Values.varnish.service.port }}
               {{- else }}
-              serviceName: {{ $fullName }}-nginx
-              servicePort: {{ $.Values.nginx.service.port }}
-              {{- end }}
+                serviceName: {{ $fullName }}-nginx
+                servicePort: {{ $.Values.nginx.service.port }}
+              {{- end }}{{- end }}
         {{- end }}
   {{- end }}
 {{- end }}
--- a/api/helm/api/templates/php-deployment.yaml
+++ b/api/helm/api/templates/php-deployment.yaml
@@ -24,27 +24,6 @@ spec:
       serviceAccountName: {{ include "api.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      initContainers:
-        - name: init-php
-          image: "{{ .Values.php.image.repository }}:{{ .Values.php.image.tag | default .Chart.AppVersion }}"
-          command: ['/bin/sh', '-c']
-          args: ['bin/console doctrine:migrations:migrate --no-interaction']
-          env:
-            - name: APP_DEBUG
-              valueFrom:
-                configMapKeyRef:
-                  name: {{ $fullName }}
-                  key: debug
-            - name: APP_ENV
-              valueFrom:
-                configMapKeyRef:
-                  name: {{ $fullName }}
-                  key: env
-            - name: DATABASE_URL
-              valueFrom:
-                secretKeyRef:
-                  name: {{ $fullName }}
-                  key: database-url
       containers:
         - name: {{ .Chart.Name }}-{{ $name }}
           image: "{{ .Values.php.image.repository }}:{{ .Values.php.image.tag | default .Chart.AppVersion }}"
--- a/api/helm/api/templates/tests/test-connection.yaml
+++ b/api/helm/api/templates/tests/test-connection.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: $fullName-test-connection
+  name: {{ $fullName }}-test-connection
   labels:
     {{- include "api.labels" . | nindent 4 }}
   annotations:
--- a/api/helm/api/values.yaml
+++ b/api/helm/api/values.yaml
@@ -4,7 +4,7 @@
 
 php:
   image:
-    repository: quay.io/api-platform/php
+    repository: "quay.io/api-platform/php"
     tag: latest
     pullPolicy: Always
   replicaCount: 1
@@ -25,7 +25,7 @@ php:
 
 nginx:
   image:
-    repository: quay.io/api-platform/nginx
+    repository: "quay.io/api-platform/nginx"
     tag: latest
     pullPolicy: Always
   replicaCount: 1
@@ -37,7 +37,7 @@ varnish:
   enabled: false
   #url: https://example.com
   image:
-    repository: quay.io/api-platform/varnish
+    repository: "quay.io/api-platform/varnish"
     tag: latest
     pullPolicy: Always
   replicaCount: 1
@@ -48,7 +48,6 @@ varnish:
 # Full configuration: https://github.com/bitnami/charts/tree/master/bitnami/postgresql
 postgresql:
   enabled: true
-  imageTag: 12-alpine
   # If bringing your own PostgreSQL, the full uri to use
   url: pgsql://api-platform:!ChangeMe!@example.com/api?serverVersion=10.1
   postgresqlUsername: "example"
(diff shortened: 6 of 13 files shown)


api/helm/api/Chart.yaml
```diff
@@ -7,10 +7,14 @@ home: https://api-platform.com
...
-      version: ~8.6.0
-      repository: https://charts.bitnami.com/bitnami
+      version: ~9.3.2
+      repository: https://charts.bitnami.com/bitnami/
       condition: postgresql.enabled
     - name: mercure
       version: ~3.0.0
       repository: https://kubernetes-charts.storage.googleapis.com/
       condition: mercure.enabled
+    - name: external-dns
```
dunglas, line 17: To remove, this shouldn't be included in this chart

api/docker/nginx/conf.d/default.tmpl
```diff
@@ -7,13 +7,12 @@ server {
     }
 
     location ~ ^/index\.php(/|$) {
-        set_by_lua $upstream_host 'return os.getenv("UPSTREAM") or "php:9000"';
-        fastcgi_pass $upstream_host;
+        fastcgi_pass "${UPSTREAM}";
         resolver local=on;
 
-        # Increase the buffer size to handle large cache invalidation headers
+        # Bigger buffer size to handle cache invalidation headers expansion
```
dunglas, line 13: Why this change? Shouldn't it be reverted?

api/helm/api/templates/php-deployment.yaml
```diff
@@ -100,23 +100,19 @@ spec:
...
                 secretKeyRef:
                   name: {{ $fullName }}
-                  key: mercure-jwt-token
+                  key: mercure-jwt-secret
           ports:
             - containerPort: 9000
           readinessProbe:
             tcpSocket:
               port: 9000
-            initialDelaySeconds: 120
```
dunglas, line 113: Why?

api/helm/api/templates/tests/test-connection.yaml
```diff
@@ -1,8 +1,9 @@
+{{- $name := "test-connection" -}}
 {{- $fullName := include "api.fullname" . -}}
 apiVersion: v1
 kind: Pod
 metadata:
-  name: $fullName-test-connection
```
dunglas, line 5: What's the benefit of this change?

api/helm/api/templates/php-deployment.yaml
```diff
@@ -100,23 +100,19 @@ spec:
                 configMapKeyRef:
                   name: {{ $fullName }}
                   key: mercure-subscribe-url
-            - name: MERCURE_JWT_TOKEN
```
dunglas, line 103: This change look buggy to me. The Symfony environment variable is `MERCURE_JWT_TOKEN`: https://github.com/symfony/recipes/blob/master/symfony/mercure-bundle/0.1/manifest.json#L12 This must be reverted.

api/helm/api/values.yaml
```diff
@@ -4,18 +4,32 @@
...
+    jwtSecret: ""
   env: prod
   debug: '0'
   secret: ""
-  corsAllowOrigin: "^https?://.*?\\.example\\.com$"
-  trustedHosts: "^.*\\.example\\.com$"
+  corsAllowOrigin: "^https?://.*\\.api-platform\\.com$"
+  trustedHosts: "^.*\\.api\\-platform\\.com$"
   trustedProxies:
+    - 103.21.244.0/22
```
dunglas, line 19: I guess that they are Cloudflare's IPs? This should be removed.

api/helm/api/values.yaml
```diff
@@ -4,18 +4,32 @@
...
   replicaCount: 1
   mercure:
-    jwtToken: ""
+    jwtSecret: ""
   env: prod
   debug: '0'
   secret: ""
-  corsAllowOrigin: "^https?://.*?\\.example\\.com$"
-  trustedHosts: "^.*\\.example\\.com$"
+  corsAllowOrigin: "^https?://.*\\.api-platform\\.com$"
```
dunglas, line 16: this must be reverted, same next line.

api/helm/api/values.yaml
```diff
@@ -34,45 +48,51 @@ nginx:
...
 
 # Full configuration: https://github.com/bitnami/charts/tree/master/bitnami/postgresql
 postgresql:
   enabled: true
-  imageTag: 12-alpine
   # If bringing your own PostgreSQL, the full uri to use
-  url: pgsql://api-platform:!ChangeMe!@example.com/api?serverVersion=10.1
-  postgresqlUsername: "example"
-  postgresqlPassword: "!ChangeMe!"
+  #url: pgsql://api-platform:!ChangeMe!@example.com/api?serverVersion=10
```
dunglas, line 66: These changes must be reverted.

api/helm/api/values.yaml
```diff
@@ -34,45 +48,51 @@ nginx:
     port: 80
 
 varnish:
-  enabled: false
+  enabled: true
```
dunglas, line 51: I wouldn't enable varnish by default.

api/helm/api/values.yaml
```diff
@@ -34,45 +48,51 @@ nginx:
...
+  image:
+    repository: bitnami/postgresql
+    tag: 10
 
 # Full configuration: https://github.com/helm/charts/tree/master/stable/mercure
 mercure:
   enabled: true
   publishUrl: http://mercure/.well-known/mercure
-  subscribeUrl: https://mercure.example.com/.well-known/mercure
+  subscribeUrl: https://mercure.api-platform.com/.well-known/mercure
```
dunglas, line 83: to revert

api/helm/api/values.yaml
```diff
@@ -34,45 +48,51 @@ nginx:
...
   allowAnonymous: "1"
-  corsAllowedOrigins: "^https?://.*?\\.example\\.com$"
+  corsAllowedOrigins: "^https?://.*\\.api-platform\\.com$"
```
dunglas, line 85: to revert

api/helm/api/values.yaml
```diff
@@ -103,13 +123,17 @@ securityContext: {}
   # runAsUser: 1000
 
 ingress:
-  enabled: false
+  enabled: true
```
dunglas, line 126: to revert, this must not be enabled by default

api/helm/api/values.yaml
```diff
@@ -103,13 +123,17 @@ securityContext: {}
...
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
   hosts:
-    - host: api.example.com
-      paths: []
+    - host: api.api-platform.com
```
dunglas, line 131: to revert

api/helm/api/templates/php-deployment.yaml
```diff
@@ -100,23 +100,19 @@ spec:
...
             - containerPort: 9000
           readinessProbe:
             tcpSocket:
               port: 9000
-            initialDelaySeconds: 120
-            periodSeconds: 3
           livenessProbe:
             tcpSocket:
               port: 9000
-            initialDelaySeconds: 120
```
dunglas, line 118: same, probably to revert

api/helm/api/values.yaml
```diff
@@ -4,18 +4,32 @@
...
 php:
   image:
-    repository: quay.io/api-platform/php
+    repository: "quay.io/api-platform/php"
     tag: latest
     pullPolicy: Always
   replicaCount: 1
   mercure:
-    jwtToken: ""
+    jwtSecret: ""
```
dunglas, line 12: to revert

api/docker/nginx/conf.d/default.tmpl
```diff
@@ -7,13 +7,12 @@ server {
     }
 
     location ~ ^/index\.php(/|$) {
-        set_by_lua $upstream_host 'return os.getenv("UPSTREAM") or "php:9000"';
-        fastcgi_pass $upstream_host;
+        fastcgi_pass "${UPSTREAM}";
```
dunglas, line 10: This will break dynamic resolution, you should use something like:

```suggestion
        set $upstream "${UPSTREAM}";
        fastcgi_pass $upstream;
```

api/helm/api/values.yaml
```diff
@@ -121,11 +145,11 @@ resources: {}
   # resources, such as Minikube. If you do want to specify resources, uncomment the following
   # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
   # limits:
-  #   cpu: 100m
-  #   memory: 128Mi
+  #  cpu: 100m
```
dunglas, line 148: probably to revert

dunglas (changes requested):

api/helm/api/templates/tests/test-connection.yaml
vincentchalamon: Without it, the variable `$fullName` seems not properly read

api/helm/api/templates/ingress.yaml
```diff
@@ -27,19 +27,23 @@ spec:
...
+    - host: {{ $host | quote }}
       http:
         paths:
         {{- range .paths }}
           - path: {{ . }}
             backend:
-              {{- if $.Values.varnish.enabled }}
-              serviceName: {{ $fullName }}-varnish
-              servicePort: {{ $.Values.varnish.service.port }}
+              {{- if contains "mercure" $host }}
```
dunglas, line 37: I'm not a fond of this. Couldn't we change the value instead to have something like the following instead of relying on the hostname?

```
ingresses:
   api: ...
   mercure: ...
```

api/helm/api/templates/php-deployment.yaml
```diff
@@ -111,12 +111,10 @@ spec:
             tcpSocket:
               port: 9000
             initialDelaySeconds: 120
-            periodSeconds: 3
```
dunglas, line 114: This should be kept.

api/helm/api/templates/php-deployment.yaml
```diff
@@ -111,12 +111,10 @@ spec:
...
           livenessProbe:
             tcpSocket:
               port: 9000
             initialDelaySeconds: 120
-            periodSeconds: 3
```
dunglas, line 119: This should be kept.

api/helm/api/templates/tests/test-connection.yaml
```diff
@@ -1,8 +1,9 @@
+{{- $name := "test-connection" -}}
```
dunglas, line 1: This variable is unused isn't it? So this could be reverted.

api/helm/api/values.yaml
```diff
@@ -37,18 +37,17 @@ varnish:
...
   #url: https://example.com
   image:
-    repository: quay.io/api-platform/varnish
+    repository: "quay.io/api-platform/varnish"
     tag: latest
     pullPolicy: Always
   replicaCount: 1
   service:
-    type: ClusterIP
+    type: NodePort
```
dunglas, line 45: I would keep a `ClusterIP` by default for security. Exposing a service from the outside of the cluster must be a conscious choice from the user.

api/helm/api/values.yaml
```diff
@@ -58,21 +57,28 @@ postgresql:
...
 mercure:
   enabled: true
   publishUrl: http://mercure/.well-known/mercure
   subscribeUrl: https://mercure.example.com/.well-known/mercure
   allowAnonymous: "1"
-  corsAllowedOrigins: "^https?://.*?\\.example\\.com$"
+  corsAllowedOrigins: "^https?://.*\\.example\\.com$"
   jwtKey: ""
+  service:
+    type: NodePort
```
dunglas, line 74: Same here, a ClusterIP should be used by default.

dunglas (changes requested):

admin/Dockerfile
```diff
@@ -28,6 +28,7 @@ COPY . ./
 VOLUME /usr/src/admin/node_modules
 
 ENV HTTPS true
+EXPOSE 3000
```
dunglas, line 31: Why is this necessary?

vincentchalamon: I tried locally with nginx-proxy. If the port is not exposed, nginx-proxy cannot map to this container

dunglas: I'm not sure to follow. This container isn't used in production.

api/helm/api/templates/ingress.yaml
vincentchalamon: I removed it cause it should be the choice of the developer to use mercure in this ingress, or directly use the mercure ingress, or another alternative.

api/helm/api/templates/configmap.yaml
```diff
@@ -8,8 +9,8 @@ data:
   env: {{ .Values.php.env | quote }}
   debug: {{ .Values.php.debug | quote }}
   cors-allow-origin: {{ .Values.php.corsAllowOrigin | quote }}
-  varnish-url: {{ if .Values.varnish.url }}{{ .Values.varnish.url | quote }}{{ else }}http://varnish{{ end }}
+  varnish-url: {{ if .Values.varnish.url }}{{ .Values.varnish.url | quote }}{{ else }}{{ printf "http://%s-varnish" $fullName }}{{ end }}
   trusted-hosts: {{ .Values.php.trustedHosts | quote }}
   trusted-proxies: {{ join "," .Values.php.trustedProxies }}
-  mercure-publish-url: {{ .Values.mercure.publishUrl | quote }}
-  mercure-subscribe-url: {{ .Values.mercure.subscribeUrl | quote }}
+  mercure-publish-url: {{ printf "http://%s-mercure/.well-known/mercure" .Release.Name | quote }}
```
soyuka, line 15: Sure this isn't `https`?

vincentchalamon: Well, it seems to work this way on api-platform/demo, so I would say yes

soyuka:
thanks @vincentchalamon

PR #1719, opened 2020-12-13
 feat: add a new Helm chart 

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      |  no 
| New feature?  | yes 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | n/a
| License       | MIT
| Doc PR        | todo

This PR adds a new Helm chart for Kubernetes.
Support for Caddy and for the Next.js container has been added.
Also, the web server now connects to PHP FPM using a Unix socket instead of the network for better performance.

* [ ] Allow configuring resources and security options per container
* [ ] Allow storing the Mercure DB in a volume


pwa/Dockerfile
```diff
@@ -12,13 +12,11 @@ EXPOSE 3000
 
 WORKDIR /usr/src/pwa
 
-RUN yarn global add @api-platform/client-generator
+ENV NEXT_TELEMETRY_DISABLED 1
```
dunglas, line 15: Should we really keep this? I don't think it's fair to the Next.js team. Or maybe could this variable default to `0`.

helm/api-platform/values.yaml
```diff
@@ -0,0 +1,133 @@
...
+    repository: "chart-example.local/api-platform/php"
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: ""
+  replicaCount: 1
+  appEnv: prod
+  appDebug: "0"
+  appSecret: ""
+  corsAllowOrigin: "^https?://.*?\\.chart-example\\.local$"
+  trustedHosts: "^127\\.0\\.0\\.1|localhost|.*\\.chart-example\\.local$"
```
alanpoulain, line 16: To change?

pwa/Dockerfile
alanpoulain: I don't have a strong opinion about it.

helm/api-platform/templates/ingress.yaml
```diff
@@ -0,0 +1,41 @@
...
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "api-platform.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if .Values.ingress.tls }}
```
VeryStrongFingers, line 19: Would be amazing if this supported the new `ingressClassName` property, for clusters that have multiple ingress controllers.

https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#setting-a-default-ingressclass

dunglas: For now I just used the file generated by `Helm create`. I try to keep the chart as much as possible un sync with the default generated chart. Do you know if Helm plans to add support for this?

VeryStrongFingers: @dunglas not quite sure. Couldn't find any related PR or open issues relating. I'll create one soon and we can update this chart later

PR #1737, opened 2020-12-31
Fix deprecated http_cache.vary

divine:
| Q             | A
| ------------- | ---
| Bug fix?      | no (2.6)
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

See @ https://github.com/api-platform/core/blob/master/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php#L370

Also, @dunglas I have a couple of questions:
- Mercure within Caddy doesn't have cors configuration and throws errors due to invalid configuration (should I add one within this template?).
- Mercure heartbeat default configuration doesn't poll every 40s to be compatible with event-source-polyfill library, adding heartbeat 30s to Caddy configuration fixes it.

Happy new year!

### DIFF
--- a/api/config/packages/api_platform.yaml
+++ b/api/config/packages/api_platform.yaml
@@ -7,12 +7,12 @@ api_platform:
         json: ['application/merge-patch+json']
     swagger:
         versions: [3]
-    # Good cache defaults
-    http_cache:
-        vary: ['Content-Type', 'Authorization', 'Origin']
     # Mercure integration, remove if unwanted
     mercure:
         hub_url: '%env(MERCURE_SUBSCRIBE_URL)%'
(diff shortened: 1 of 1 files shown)


dunglas:
Thank you!!

Regarding your questions:

1. It's on purpose, as the API, the PWA and Mercure are now served on the same domain, CORS settings aren't mandatory anymore. It should be configurable using the `MERCURE_EXTRA_DORECTIVES` environment variable
2. It's not normal, it should the default. It's probably a bug in Mercure itself. I'll take a look.

dunglas (approved):

Issue #1726, opened 2020-12-21
next/link not works


nicotec:
**API Platform version(s) affected**: 2.6.beta (wsl2)

**Description**  
If we add a link with next / link. An error always occurs. The code is the original one.

**How to reproduce**  
import Head from "next/head";
import Link from "next/link";

const Welcome = () => (
  <>
    <Head>
      <title>Welcome to API Platform!</title>
    </Head>

    <div className="welcome">
      <Link href="/"><a>test link</a></Link>
      <header className="welcome__top">

**Possible Solution**  
https://github.com/vercel/next.js/issues/16874

**Additional Context**  
![image](https://user-images.githubusercontent.com/10002262/102797861-8cc5a400-43b0-11eb-8441-938068686241.png)

LeoBenoist:
Hello. same problem for me.

It is related to the caddy Caddyfile.

Next json request are forwarded to Symfony instead of Next due to:

api/docker/caddy/Caddyfile

`@pwa {
    # Pass the request to the API for typical API media types.
    # `application/xml` is not in the list on purpose:
    # it is part of the default Accept header sent by browsers.
    not header_regexp Accept (?i)(?:\bjson\b|^text/(?:csv|xml)$|^application/x-yaml$)

    # Explicitly list routes always handled by the API as well as
    # URLs with an extension corresponding to common formats supported by the API.
    not path_regexp (?i)(?:^/docs|^/graphql|^/bundles/|^/_profiler|^/_wdt|.+\.(?:json|html$|csv$|ya?ml$|xml$))
}`

dunglas:
Good catch. We must add an exception for the `/_next` prefix in the regex.

LeoBenoist:
Also manifest.json is missing if have next pwa

LeoBenoist:
For now I have fixed my project like that: (probably some better ways to it but I don't know caddy

```
{
    # Debug
    {$DEBUG}
    # HTTP/3 support
    servers {
        protocol {
            experimental_http3
        }
    }
}

{$SERVER_NAME}

log

@pwaNext {
    path /_next/*
    path /manifest.json
}

@pwa {
    # Pass the request to the API for typical API media types.
    # `application/xml` is not in the list on purpose:
    # it is part of the default Accept header sent by browsers.
    not header_regexp Accept (?i)(?:\bjson\b|^text/(?:csv|xml)$|^application/x-yaml$)

    # Explicitly list routes always handled by the API as well as
[...]

PR #1741, opened 2021-01-03
fix: Caddyfile and default JS deps for the PWA

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes https://github.com/api-platform/api-platform/issues/1726
| License       | MIT
| Doc PR        | todo

### DIFF
--- a/pwa/pages/index.js
+++ b/pwa/pages/index.js
@@ -1,4 +1,5 @@
 import Head from "next/head";
+import Link from "next/link";
 
 const Welcome = () => (
   <>
@@ -58,7 +59,7 @@ const Welcome = () => (
               </div>
               <div className="other__content">
                 <h3>
-                  <a href="/docs">API</a>
+                  <Link href="/docs">API</Link>
                 </h3>
               </div>
             </div>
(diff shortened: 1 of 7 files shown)


LeoBenoist:
@dunglas this is working for me for next. 
⚠️ I don't know if you want to support by default PWA.

For PWA it would require to add an exception for the  manifest.json `|| {path}.startsWith("/manifest.json")` It is the most commonly used URL in documentation around the web (including Mozzila and Google).

dunglas:
Good catch, I'll add it!

LeoBenoist:
You’re welcome :). Thank you very much for your work and API platform. On my side it will power a small startup that has for goal to make the world a better place in Africa.

pwa/pages/index.js
```diff
@@ -58,7 +59,7 @@ const Welcome = () => (
               </div>
               <div className="other__content">
                 <h3>
-                  <a href="/docs">API</a>
+                  <Link href="/docs">API</Link>
```
LeoBenoist, line 62: A small one here. With next/link I think it should be `<Link href="/docs"><a>API<a/a></Link>`. Link is only a wrapper on next contrary to react router. (false friend)

PR #1752, opened 2021-01-09
Force composer 2.0 for building the php container

cedricziel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes 
| New feature?  | no 
| BC breaks?    | no
| Deprecations? | no
| Tickets       | fixes #... 
| License       | MIT

Using composer:latest as `COPY --from` reference imposes the risk of colliding
with an already existing `composer:1.x` container. In that case, the `1.x` container will
be re-used. It will fail upon start with an unknown named parameter exception

This change pins the composer image in use to the `2.0` tag, meaning it will either
use an existing 2.0 container or pull the latest one.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -6,7 +6,6 @@
 # https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG PHP_VERSION=8.0
 ARG CADDY_VERSION=2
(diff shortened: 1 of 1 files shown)


dunglas (approved):

dunglas:
Thank you!

PR #1823, opened 2021-02-23
Add headers for a more secure PWA.

johnhout:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes/no
| New feature?  | no
| Deprecations? | no
| Tickets       |   Ran into this issue on my own instead of adding an issue added a pull request.
| License       | MIT

Implement a few headers for the PWA to protect the application against Clickjacking and XSS attacks based on the OWASP guidelines.

### DIFF
--- a/api/docker/caddy/Caddyfile
+++ b/api/docker/caddy/Caddyfile
@@ -44,10 +44,14 @@ route {
     vulcain
     push
 
-    # Set headers for protection against XSS, clickjacking and MIME-sniffing.
-    header / {
+    header {
+        # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
+        ?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
+
+        # Set security headers
+
         # Enable HTTP Strict Transport Security (HSTS) to force clients to always connect via HTTPS
-        Strict-Transport-Security "max-age=31536000;"
+        Strict-Transport-Security "max-age=31536000; preload"
         # Enable cross-site filter (XSS) and tell browser to block detected attacks
         X-XSS-Protection "1; mode=block"
         # Prevent some browsers from MIME-sniffing a response away from the declared Content-Type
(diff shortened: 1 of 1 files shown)


api/docker/caddy/Caddyfile
```diff
@@ -47,6 +47,18 @@ route {
     # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
     header ?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
 
+    # Set headers for protection against XSS, clickjacking and MIME-sniffing.
```
dunglas, line 50: Couldn't this block be merged with the previous line defining the `Link` header?

johnhout: Agree, will update the pull request

dunglas: I mean can't we move `?Link `</docs.jsonld>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation", </.well-known/mercure>; rel="mercure"`
` inside the `header / {` block?

dunglas:
@johnhout I updated your PR, but [according to OWASP](https://owasp.org/www-project-secure-headers/) the `X-XSS-Protection` header is deprecated and should be replaced by a Content Security Policy, and the `Referrer-Policy` should be set. WDYT?

francislavoie:
This will need to be reorganized once https://github.com/api-platform/api-platform/pull/1893 is merged -- it's not correct to have `?` header defaults in the same `header` handler as other rules.

dunglas:
We should use the Next.js config instead, so there will be set even if the user doesn't use Caddy: https://nextjs.org/docs/advanced-features/security-headers

rvanlaak:
I guess this PR was superseded by the merge of #1893 ?

francislavoie:
My PR did something different than this @rvanlaak it was fixing a config bug related to `?` conditional header setting.

I don't think this PR is needed though, those additional "security headers" should be set in the API Platform app itself (via a middleware) as needed and not set in the web server. Those security headers are an application-layer concern, not a webserver-layer.

PR #1844, opened 2021-03-08
Rename env variable from REACT_APP_API_ENTRYPOINT to NEXT_PUBLIC_REAC…

### LINKED ISSUE #1843: Default PWA environment variables won't work in production
**API Platform version(s) affected**: 2.6.3

**Description**  
The default PWA environment variables will never work in production because as stated in the Next.js documentation : 

> In order to expose a variable to the browser you have to prefix the variable with `NEXT_PUBLIC_`

The following line is used to define the API URL :
https://github.com/api-platform/api-platform/blob/main/pwa/.env#L1

As it's not prefixed with `NEXT_PUBLIC_` so when you try to use it in production, you get calls to http://localhost
I wasn't aware of this rule and searched for a while before understanding what was going on.

**Possible Solution**  
Rename REACT_APP_API_ENTRYPOINT to NEXT_PUBLIC_REACT_APP_API_ENTRYPOINT

Oryss:
…T_APP_API_ENTRYPOINT (#1843)

| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #1843
| License       | MIT
| Doc PR        |

### DIFF
--- a/pwa/.env
+++ b/pwa/.env
@@ -1,1 +1,1 @@
-REACT_APP_API_ENTRYPOINT=https://localhost
+NEXT_PUBLIC_REACT_APP_API_ENTRYPOINT=https://localhost
--- a/pwa/Dockerfile
+++ b/pwa/Dockerfile
@@ -38,7 +38,7 @@ CMD ["yarn", "dev"]
 FROM api_platform_pwa_common AS api_platform_pwa_prod
 
 ENV NODE_ENV production
-ARG REACT_APP_API_ENTRYPOINT
+ARG NEXT_PUBLIC_REACT_APP_API_ENTRYPOINT
 
 RUN set -eux; \
 	yarn build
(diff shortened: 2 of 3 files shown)


dunglas (approved):

dunglas:
After a second thought we can now just remove this environment variable and use `window.origin` in `index.ts` as we have done in https://github.com/api-platform/api-platform/blob/main/pwa/config/entrypoint.ts.

WDYT @Oryss?

Oryss:
> After a second thought we can now just remove this environment variable and use `window.origin` in `index.ts` as we have done in https://github.com/api-platform/api-platform/blob/main/pwa/config/entrypoint.ts.
> 
> WDYT @Oryss?

I agree. I wasn't aware you could use this variable and http://caddy to get the entrypoint. It works fine!

dunglas:
Do you want to open a new PR with this change? If you have no time for this I can do it.

PR #1887, opened 2021-04-20
Add some updates

vincentchalamon:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Tickets       | N/A
| License       | MIT
| Doc PR        | N/A

> [WARNING] templates/ingress.yaml: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

Cluster api-platform-demo has been upgraded to 1.19 (see [release notes](https://cloud.google.com/kubernetes-engine/docs/release-notes)).

Note: `yarn eslint` not available because of https://stackoverflow.com/questions/59633005/how-is-eslint-integrated-into-create-react-app?answertab=votes#tab-top


.github/workflows/ci.yml
```diff
@@ -35,5 +36,34 @@ jobs:
               run: curl  -vk -o /dev/null https://localhost
             - name: PHPUnit
               run: docker-compose exec -T php bin/phpunit
+            - name: Doctrine Schema Validator
+              run: docker-compose exec -T php bin/console doctrine:schema:validate --skip-sync
             - name: Jest
-              run: docker-compose exec -T pwa yarn test --passWithNoTests
+              run: docker-compose exec -T pwa yarn test --ci --passWithNoTests
+
+    security-check:
```
dunglas, line 44: GitHub does this automatically, IMHO this isn't needed because this is only works on GitHub anyway.

api/config/services_test.yaml
```diff
@@ -0,0 +1,6 @@
+services:
```
dunglas, line 1: Shouldn't this be added in the Flex recipe instead? We synchronize the config files with the official recipes before every release.

.github/workflows/ci.yml
vincentchalamon: I was wondering about it too, using dependabot, but couldn't find any article about it.

api/public/robots.txt
```diff
@@ -0,0 +1,5 @@
+# www.robotstxt.org/
```
dunglas, line 1: What's the point of adding this file?

.github/workflows/ci.yml
dunglas: It's documented here: https://docs.github.com/en/code-security/supply-chain-security/about-alerts-for-vulnerable-dependencies

api/.gitignore
```diff
@@ -1,7 +1,3 @@
-/docker/db/data
```
dunglas, line 1: These deletions should be reverted isn't it?

docker-compose.yml
```diff
@@ -2,6 +2,7 @@ version: "3.4"
 
 services:
   php:
+    image: ${CONTAINER_REGISTRY_BASE:-ghcr.io/api-platform/api-platform}/php:${CONTAINER_REGISTRY_TAG:-latest}
```
dunglas, line 5: This must be removed. We don't upload pre-built images to a registry anymore (it is not worth the complexity).

api/.env.test
```diff
@@ -6,4 +6,4 @@ PANTHER_APP_ENV=panther
 PANTHER_ERROR_SCREENSHOT_DIR=./var/error-screenshots
 
 # API Platform distribution
-TRUSTED_HOSTS=^example.com$
+TRUSTED_HOSTS=^example\.com|localhost$
```
dunglas, line 9: why this change?

.github/workflows/ci.yml
```diff
@@ -1,39 +1,54 @@
 name: CI
 
 on:
-    push:
-    pull_request:
+  push:
```
dunglas, line 4: Can you preserve the indentation please? IIRC it's in sync with GitHub's best practices.

dunglas (changes requested):

docker-compose.yml
vincentchalamon: But how can we name the images built from `docker-compose build` before deploy?

api/.env.test
vincentchalamon: The WebTestCase defines the client host to http://localhost, but the ApiTestCase defines it to http://example.com

.github/workflows/ci.yml
vincentchalamon: Well, the .editorconfig is wrong so: https://github.com/api-platform/api-platform/blob/main/.editorconfig#L50-L52

vincentchalamon: Indent with 2 seems already GitHub's best practices:

> In YAML examples, such as actions and workflow files, use two spaces to indent lines within nested lists and block sequences.

https://github.com/github/docs/blob/main/contributing/content-style-guide.md#indentation

docker-compose.yml
dunglas: Why do we need to use `docker-compose` to build the images for k8s?

api/Dockerfile
```diff
@@ -103,10 +103,10 @@ RUN set -eux; \
 	chmod +x bin/console; sync
 VOLUME /srv/api/var
 
-COPY docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
-RUN chmod +x /usr/local/bin/docker-healthcheck
-
-HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["docker-healthcheck"]
+ENV FCGI_CONNECT=/var/run/php/php-fpm.sock
+RUN wget -O /usr/local/bin/php-fpm-healthcheck https://raw.githubusercontent.com/renatomefi/php-fpm-healthcheck/master/php-fpm-healthcheck; \
```
dunglas, line 107: Can we copy (and adapt) this script instead of relying on an external repository please?

api/Dockerfile
```diff
@@ -103,10 +103,10 @@ RUN set -eux; \
 	chmod +x bin/console; sync
 VOLUME /srv/api/var
 
-COPY docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
-RUN chmod +x /usr/local/bin/docker-healthcheck
-
-HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["docker-healthcheck"]
+ENV FCGI_CONNECT=/var/run/php/php-fpm.sock
```
dunglas, line 106: Why do we need this env var?

docker-compose.yml
vincentchalamon: How to you recommend to build Docker images for k8s? :thinking:

api/Dockerfile
vincentchalamon: For php-fpm-healthcheck to use this sock instead of the default value `localhost:9000`

dunglas: Maybe could we hardcore this as default value in the script?

api/.env.test
vincentchalamon: https://github.com/api-platform/core/issues/4242

.github/workflows/ci.yml
```diff
@@ -1,39 +1,54 @@
...
+      - name: Jest
+        run: docker-compose exec -T pwa yarn test --ci --passWithNoTests
+
+  lint-helm:
+    name: Lint Helm
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
```
dunglas, line 52: This blank line can be removed for consistency.

api/Dockerfile
```diff
@@ -103,10 +103,9 @@ RUN set -eux; \
 	chmod +x bin/console; sync
 VOLUME /srv/api/var
 
-COPY docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
-RUN chmod +x /usr/local/bin/docker-healthcheck
-
-HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["docker-healthcheck"]
+COPY docker/php/php-fpm-healthcheck /usr/local/bin/php-fpm-healthcheck
+RUN chmod +x /usr/local/bin/php-fpm-healthcheck
+HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["php-fpm-healthcheck", "-v"]
```
dunglas, line 108: Why do we need to use the verbose mode?

api/docker/php/php-fpm-healthcheck
```diff
@@ -0,0 +1,139 @@
...
+command -v tail 1> /dev/null || { >&2 echo "Make sure tail is installed (i.e. apk add --no-cache busybox). Aborting."; exit 4; }
+command -v grep 1> /dev/null || { >&2 echo "Make sure grep is installed (i.e. apk add --no-cache grep). Aborting."; exit 4; }
+
+# Get status from fastcgi connection
+# $1 - cgi-fcgi connect argument
+get_fpm_status() {
+    if test "$VERBOSE" = 1; then printf "Trying to connect to php-fpm via: %s%s\\n" "$1" "$SCRIPT_NAME"; fi;
+    
+    # Since I cannot use pipefail I'll just split these in two commands
+    FPM_STATUS=$(env -i REQUEST_METHOD="$REQUEST_METHOD" SCRIPT_NAME="$SCRIPT_NAME" SCRIPT_FILENAME="$SCRIPT_FILENAME" "$FCGI_CMD_PATH" -bind -connect "$1" 2> /dev/null)
```
dunglas, line 58: Isn't this doing basically the same thing as the script we currently have? I don't understand why our script fails but not this one.

helm/api-platform/templates/deployment.yaml
```diff
@@ -130,13 +130,15 @@ spec:
           readinessProbe:
             exec:
               command:
-                - docker-healthcheck
+                - php-fpm-healthcheck
+                - -v
```
dunglas, line 134: Same here, do we need the verbose output?

dunglas (approved):

api/docker/php/php-fpm-healthcheck
vincentchalamon: Found. It's fixed in this PR (rollbacked to docker-healthcheck)

api/Dockerfile
```diff
@@ -105,7 +105,6 @@ VOLUME /srv/api/var
 
 COPY docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
 RUN chmod +x /usr/local/bin/docker-healthcheck
-
```
dunglas, line 108: This could maybe be reverted to reduce the diff.

api/config/packages/framework.yaml
```diff
@@ -2,7 +2,6 @@
 framework:
     secret: '%env(APP_SECRET)%'
     trusted_proxies: '%env(TRUSTED_PROXIES)%'
-    trusted_hosts: '%env(TRUSTED_HOSTS)%'
```
dunglas, line 5: This must be reverted.

helm/api-platform/templates/ingress.yaml
```diff
@@ -1,10 +1,11 @@
 {{- if .Values.ingress.enabled -}}
```
dunglas, line 1: We should use @sagikazarmark's definition instead: https://github.com/sagikazarmark/helm-charts/blob/master/charts/caddy/templates/ingress.yaml

pwa/Dockerfile
```diff
@@ -38,6 +38,7 @@ CMD ["yarn", "dev"]
 FROM api_platform_pwa_common AS api_platform_pwa_prod
 
 ENV NODE_ENV production
+ARG ENTRYPOINT
```
dunglas, line 41: AFAIU, it's useless.

pwa/config/entrypoint.ts
```diff
@@ -1 +1 @@
-export const ENTRYPOINT = typeof window === "undefined" ? "http://caddy" : window.origin;
+export const ENTRYPOINT = typeof window === "undefined" ? process.env.ENTRYPOINT : window.origin;
```
dunglas, line 1: We should add a default value for the `ENTRYPOINT` env var here or in the `docker-compose.yaml`.

dunglas (changes requested):

dunglas (approved):

dunglas:
Thanks @vincentchalamon! Great work.

PR #1893, opened 2021-04-23
fix: header defaults in Caddyfile must be in separate handlers

francislavoie:
Discussion in https://github.com/caddyserver/caddy/issues/4130 brought this up.

There's a bit of a gotcha here, in that having two `?Field` headers will append each to the `Require` response matcher, so those header fields will only be set if _both_ are not present in the response. If one is present (e.g. `Link` is in the response) then neither are set.

Using separate header handlers solves this because they will use their own handlers, and therefore separate response matchers.

This is just a followup to https://github.com/api-platform/api-platform/pull/1879 to fix this non-obvious behaviour.

| Q             | A
| ------------- | ---
| Branch?       | main 
| Bug fix?      | yes 
| New feature?  | no 
| Deprecations? | no
| Tickets       | N/A 
| License       | MIT
| Doc PR        | N/A

### DIFF
--- a/api/docker/caddy/Caddyfile
+++ b/api/docker/caddy/Caddyfile
@@ -44,12 +44,10 @@ route {
     vulcain
     push
 
-    header {
-        # Add links to the API docs and to the Mercure Hub if not set explicitly (e.g. the PWA)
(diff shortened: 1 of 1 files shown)


dunglas (approved):

dunglas:
Thanks @francislavoie!

PR #1938, opened 2021-06-16
feat: migrate to yarn 2

djpremier:
| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT


dunglas:
I don't think that we want to vendor the dependencies in this repository.
Also, Yarn 2 doesn't look very popular. Maybe should we just switch to NPM?

djpremier:
> I don't think that we want to vendor the dependencies in this repository.
> Also, Yarn 2 doesn't look very popular. Maybe should we just switch to NPM?

Thanks for the return @dunglas.

In the next few days I will be putting together a text with details on what I consider to be the justification for migrating. I made the pull request as a way for other people to give their opinion about it and for possible interested parties who already have an initiation.
Yes, we can use NPM or others (there are several), but the fact that Yarn 2 is not so popular I believe that it is precisely because of this fact of the dependencies that people end up seeing it as something bad, but in the company's front-end project which work actually helped, because in addition to maintaining consistency (one of the topics that the Yarn team points out) it also reduces time, a Docker build decreases by up to 5 times, while the increase in the repository was negligible (few kilobytes). Clearly when updating too many dependencies the commit gets "a little polluted", but it has its advantages in other aspects.

divine:
> I don't think that we want to vendor the dependencies in this repository.
> Also, Yarn 2 doesn't look very popular. Maybe should we just switch to NPM?

Yarn 2 isn't backward compatible and breaks a lot of projects so it can't be just migrated, it requires complete migration and intervention.

Better to keep away from Yarn 2 and keep things simple.

> Yarn 2 is not so popular I believe that it is precisely because of this fact of the dependencies that people end up seeing it as something bad

It's not popular because of drastic change and it made only things messy. Nobody is moving to Yarn 2 and that's why Yarn developers are working on bringing back-compatible version.

Thanks!

rvanlaak:
Would pnpm be an option?

dunglas:
We switched to `pnpm`. Thanks for your contrib!

PR #1939, opened 2021-06-16
fix: export ENTRYPOINT in pwa/config/entrypoint.ts

alanpoulain:
The `dataAccess` module imports the `ENTRYPOINT` const, and not `NEXT_PUBLIC_ENTRYPOINT`:
https://github.com/api-platform/client-generator/blob/cf9663004bb839965c49b4db2de6d18bea696f74/templates/next/utils/dataAccess.ts#L5


dunglas:
This should be changed in the client generator IMHO. We need this convention to have access to the env var when using SSG.

alanpoulain:
I'm not changing the env var name. Only the exported const.

alanpoulain:
You can see it fully working here: https://demo-fix-export-entrypoint.api-platform.com

dunglas (approved):

dunglas:
Thanks Alan!

Issue #1859, opened 2021-03-24
Deploy api-platform 2.6 with docker


d0niek:
Hi!

Deploy with docker to version **2.5** was easy with https://github.com/api-platform/docker-compose-prod

On local:

```
vim .env
docker-compose -f docker-compose.build.yml build --pull
docker-compose -f docker-compose.build.yml push
```

On server:

```
vim .env
docker-compose pull
docker-compose up -d
docker-compose exec php bin/console cache:clear && docker-compose restart php
```

How should I do it with version **2.6**? Which **docker-compose\*** file(s) use for build/push, which **docker-compose\*** file(s) use for pull/up?

For example file **docker-compose.override.yml** expose caddy ports so it looks good for `docker-compose up -d` but it uses build target **dev** for pwa and bind local directory for php.

d0niek:
Ok, I see there was an update in **2.6.3** version and now I think it should looks like:

On local:

```
docker-compose -f docker-compose.yml -f docker-compose.prod.yml build --pull
docker-compose -f docker-compose.yml -f docker-compose.prod.yml push
```

On server:

```
docker-compose -f docker-compose.yml -f docker-compose.prod.yml pull
docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d
```

Am I correct? :-)

jlagneau:
Yes, that's the correct way to deploy.

Although, you don't need to specify docker-compose.override.yml for development, you can just use docker-compose normally, read docker-compose docs.

https://docs.docker.com/compose/production/
https://docs.docker.com/compose/extends/

> By default, Compose reads two files, a `docker-compose.yml` and an optional `docker-compose.override.yml` file. By convention, the `docker-compose.yml` contains your base configuration. The override file, as its name implies, can contain configuration overrides for existing services or entirely new services.

In this case, the override file contains the development configuration.

The process for deploying in production in 2.5 was slightly more complicated than that, there was an external repository for production [api-platform/docker-compose-prod](https://github.com/api-platform/docker-compose-prod) but it was archived and is no longer relevant.

PR #1965, opened 2021-07-08
Simplify deploy with docker-compose

d0niek:
According to this issue #1859 to run docker-compose on production we
have to run command like
`docker-compose -f docker-compose.yml -f docker-compose.prod.yml ...`.

With setting up an env COMPOSE_FILE we don't have to prefix docker
commands with compose files.

See https://docs.docker.com/compose/reference/envvars/#compose_file


.env.dist
```diff
@@ -0,0 +1,2 @@
+# Uncomment line below on production server to skip docker-compose -f options
```
dunglas, line 1: Is the `.env.dist` file automatically loaded? I cannot find any information regarding this on https://docs.docker.com/compose/env-file/

.env.dist
```diff
@@ -0,0 +1,2 @@
...
+#COMPOSE_FILE=docker-compose.yml:docker-compose.prod.yml
```
dunglas, line 2: We should also add a default `SERVER_NAME` env var here to hint users that they can use this env var to configure the hostname (cc @francislavoie).

.env.dist
francislavoie: It is not, typically on repo setup, `cp .env.dist .env` should be run, and `.env.dist` is basically an example that lives with the repo

dunglas: In this case shouldn't we rename this file `.env.prod` and uncomment the env var?

francislavoie: I'd say that's probably appropriate considering this project's existing setup. The word `dist` is a bit incongruent with what this repo currently has.

.env.dist
d0niek: Sounds good and we can add more envs like `APP_SECRET`, `MERCURE_PUBLISHER_JWT_KEY`, `MERCURE_SUBSCRIBER_JWT_KEY` but is it a good PR to do that?

soyuka:

Issue #2003, opened 2021-08-30
[Helm Chart] Wrong value taken for replicaCount


WhiteRabbitDE:
**Description**  
In the helm template "deployment.yaml" the wrong .Values. value is taken for replicas. The used value does not exist.

https://github.com/api-platform/api-platform/blob/7d8eb360592583a7b7ca76708c64c6d495e26383/helm/api-platform/templates/deployment.yaml#L9

It should be taken from .Values.php.replicaCount i guess or moved to the root in values.yaml. I may be wrong but right now the values in  php and caddy for replicaCount are unused and can not be scaled independently.

**Possible Solution**  
Move the replicaCount value to the root of values.yaml or use .Values.php.replicaCount

dunglas:
Good catch! Do you want to open a PR?

WhiteRabbitDE:
Sure, i can create a PR if you help me decide for which solution to use. 

Using the  .Values.php.replicaCount and remove the replicaCount under the caddy node seems good to me, wdyt ?

dunglas:
Moving it to the root looks better to me because we now have only 1 pod containing both the Caddy and the PHP containers. WDYT?

WhiteRabbitDE:
You are right ! But maybe i would move it into the autoscaling block then to have it the same way we do with the  postgres url which is used if postgres.disabled.

```  
{{- if .Values.postgresql.enabled }}
  database-url: {{ printf "pgsql://%s:%s@%s-postgresql/%s?serverVersion=13&charset=utf8" .Values.postgresql.postgresqlUsername .Values.postgresql.postgresqlPassword .Release.Name .Values.postgresql.postgresqlDatabase | b64enc | quote }}
  {{- else }}
  database-url: {{ .Values.postgresql.url | b64enc | quote }}
  {{- end }}
```
But I'm fine with moving it to the root , too.

PR #2004, opened 2021-08-30
fix: replicaCount value in helm chart

WhiteRabbitDE:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #2003
| License       | MIT
| Doc PR        | 

replicaCount always used the default value of 1 because of https://github.com/api-platform/api-platform/blob/7d8eb360592583a7b7ca76708c64c6d495e26383/helm/api-platform/templates/deployment.yaml#L9


dunglas (approved):

dunglas:
Thanks, @WhiteRabbitDE!

Issue #2025, opened 2021-09-19
Build PHP Failed with Error 127 Operation Not Permitted 


bspoved:
**API Platform version(s) affected**: x.y.z

**Description**  

when running ` #> docker-compose up -d  ` i had this error
```
make: /bin/sh: Operation not permitted
/bin/sh /usr/src/php/ext/intl/libtool --mode=compile cc -I. -I/usr/src/php/ext/intl -I/usr/src/php/ext/intl/include -I/usr/src/php/ext/intl/main -I/usr/src/php/ext/intl -I/usr/local/include/php -I/usr/local/include/php/main -I/usr/local/include/php/TSRM -I/usr/local/include/php/Zend -I/usr/local/include/php/ext -I/usr/local/include/php/ext/date/lib  -fstack-protector-strong -fpic -fpie -O2 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -DHAVE_CONFIG_H  -fstack-protector-strong -fpic -fpie -O2 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64    -DU_NO_DEFAULT_INCLUDE_UTF_HEADERS=1 -DU_HIDE_OBSOLETE_UTF_OLD_H=1 -Wno-write-strings -D__STDC_LIMIT_MACROS -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 -c /usr/src/php/ext/intl/php_intl.c -o php_intl.lo 
make: *** [Makefile:209: php_intl.lo] Error 127
[...]

PR #2026, opened 2021-09-19
update the alpine version

bspoved:
update the alpine version to 3.13  to fix the issue https://github.com/api-platform/api-platform/issues/2025

| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -8,7 +8,7 @@ ARG PHP_VERSION=8.0
 ARG CADDY_VERSION=2
 
 # "php" stage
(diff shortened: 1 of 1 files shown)


djpremier:
Yes, this error is reported here:

https://github.com/docker-library/php/issues/1130
https://github.com/alpinelinux/docker-alpine/issues/146

dunglas:
We try to avoid using fixed versions of Alpine to not increase the maintenance burden. This error should be fixed upstream.

PR #2065, opened 2021-11-09
chore: Upgrade to Next 12

theMadness:
Including opt-in SWC minification.

Babel file preserved for Jest, but excluded from Next visibility.

Can't think of any tests to add, as this either breaks the whole suite or none of it.


dunglas (approved):

dunglas:
Thank you very much @theMadness!

PR #2066, opened 2021-11-10
feat: Update apcu version in Dockerfile

WhiteRabbitDE:
Fixed compatibility with PHP 8.1 by adding return types to APCUIterator.


dunglas:
Thanks!

PR #1228, opened 2019-08-19
Add a reference Docker Compose setup for production

teohhanhui:
As an alternative to Helm / Kubernetes, for those who are uncomfortable with it. (I'm one of them :laughing:)

I've tested (on GCE) the Let's Encrypt / nginx proxy, and it seems to work fine.

Also removed Varnish ("cache-proxy" container) from the (dev) Docker Compose setup, as per https://github.com/api-platform/core/issues/1856#issuecomment-522497404

TODO:
- [ ] Docs

PR #2067, opened 2021-11-10
Do not create VOLUME for node_modules inside pwa/Dockerfile

maks-rafalko:
It was added here without any explanation https://github.com/api-platform/api-platform/pull/1228 by @teohhanhui (if you have any comments - that would be very helpful)

Notes:

* Inside `api/Dockerfile` - there is no `VOLUME` for `vendor` folder, so it's not clear why we use `VOLUME` for JS application
* Because of this `VOLUME`, node modules folder is populated only inside docker volume, but **not on the host machine** during development. This required developers to run `yarn install` locally in order to have `node_modules` populated for IDE inspections

After this update, `node_modules` will work the same way as `vendor` folder works inside `api` app, so it will be shared thanks to docker bind-mount from `docker-compose.override.yaml`:

https://github.com/api-platform/api-platform/blob/0c19b09a05c268b4326b1cf9e32fcef5c4e238fc/docker-compose.override.yml#L19

I've targeted `main` branch because it may affect developer's workflow.

To start using this approach, I had to remove volume locally and rebuild the image:

* `docker-compose down -v` (too aggressive but ok in my case)
* `docker-compose build --no-cache --force-rm client`

### DIFF
--- a/pwa/Dockerfile
+++ b/pwa/Dockerfile
@@ -20,8 +20,6 @@ RUN yarn
 
 COPY . ./
 
-VOLUME /usr/src/pwa/node_modules
-
 
 # "development" stage
 # depends on the "common" stage above


alanpoulain:
I think the volume has been added because the bind-mounted `node_modules` folder is very heavy to synchronize for Mac users and degrades the performances.
Have you tried to create it like it's done for the `var` folder (commented)?
https://github.com/api-platform/api-platform/blob/0c19b09a05c268b4326b1cf9e32fcef5c4e238fc/docker-compose.override.yml#L10

maks-rafalko:
> I think the volume has been added because the bind-mounted `node_modules` folder is very heavy to synchronize for Mac users and degrades the performances.

but now it affects all users - including those who use Linux with a native docker support.

Also, since it's located inside `Dockerfile` ([but not `docker-compose.override.yaml`](https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose)) - the image built on top of such `Dockerfile` will always create a `VOLUME` for `node_modules`, even if we don't use `docker-compose`.

From the [docs](https://docs.docker.com/engine/reference/builder/#volume):

> This `Dockerfile` results in an image that causes `docker run` to create a new mount point at /myvol and copy the greeting file into the newly created volume.

Also, look at the following chapter ["Notes about specifying volumes"](https://docs.docker.com/engine/reference/builder/#notes-about-specifying-volumes):

> **Changing the volume from within the Dockerfile:** If any build steps change the data within the volume after it has been declared, those changes will be discarded.

and, again from my understanding, this is exactly what happens here:

https://github.com/api-platform/api-platform/blob/0c19b09a05c268b4326b1cf9e32fcef5c4e238fc/pwa/Dockerfile#L40-L44

`yarn build` with `NODE_ENV=prodcution` should remove `devDependencies` from `node_modules`. But since `VOLUME` has already been created - this will not have any impact

---

TL;DR I don't think we should add anything related to mac-users to the base Dockerfile. This doesn't sound right.

Mac users can use volumes on their `docker-compose.override.yaml` files if needed, like:

```yaml
services:
  pwa:
    volumes:
      # ...
      - /usr/src/node_modules
```

---

Alternatively, if you have explanation for the concerns above and don't want to change anything, at least please consider adding comments that this behavior is for mac-users only, because now it's not clear and nobody from my team was able to understand why it's in the base Dockerfile

dunglas:
This looks legit to me. Could you just add a commented line in `docker-compose.override.yaml` allowing to create a module to improve performance on Mac and Windows?

Thanks

dunglas:
Reverted because this breaks the distribution.

maks-rafalko:
Could you please elaborate on this?

dunglas:
Check the CI run or try it locally. This breaks the image because node_modules aren't found.

Issue #797, opened 2018-08-05
Loading API gives 502 bad gateway


krishnathokala:
HI All, 

When i load the API using http://localhost:8080/ URL it gives me "502 Bad Gateway". Could anyone please guide me what is the issue and how to fix this.

Thanks in Advance.

teohhanhui:
Did you verify that PHP-FPM in the `php` container is already listening?

`docker-compose logs -f php`

vbourdeix:
Don't know if that will help, but when that happens, I docker-compose stop then docker-compose up -d my different services and generally it works. I don't like the fact that I don't really understand this issue, but it allowed me to continue working on my project without wasting time on Docker related stuff.

wuestkamp:
A fast fix for this is to call `docker-compose exec php php-fpm -D` to start php-fpm again. I did a complete rebuild of docker-compose and still getting this error.

Anyone got an idea why this happens? It does only so on my local and not my Gcloud Kubernetes cluster.

The [Dockerfile](https://github.com/api-platform/api-platform/blob/master/api/Dockerfile) should run php-fpm.

v2.3.1

wildcard27:
Was there any fix for this? I have just upgraded to 2.3.6 and am getting the same problem.

Everytime I restart the container, I need to run `docker-compose exec php php-fpm -D`. This is getting super annoying

wildcard27:
```
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response
php_1          | /tmp:5432 - no response

```
etc, etc, etc

wildcard27:
It seems that the problem is that the `api` container waits for PHP-FPM to be ready in `docker-entrypoint.sh` but is perhaps timing out?

wildcard27:
@aaa2000 Have already done that

baddlan:
There is a good chance `DATABASE_URL` is not set. Try checking your env vars inside the `php` container.

wildcard27:
@baddlan The `docker-compose.yml` doesn't have that param: https://github.com/api-platform/api-platform/blob/v2.3.6/docker-compose.yml

baddlan:
[...]

Issue #594, opened 2018-03-03
Switching to MySQL


mvaljento:
I'm a newbie to Docker and Api Platform. I got the latest version running on my Mac, but I'm having trouble changing over from Postgres to MySQL and I did not find anything on it in documentation. 

MySQL seems to be running, but I cannot run php commands to for example create the schema.

Any idea where changes need to be made?

From my docker-compose.yaml:
```

image: mysql
        volumes:
            - "./.data/mysql:/var/lib/mysql"
        environment:
            MYSQL_USER: lorem
            MYSQL_ROOT_PASSWORD: lorem
            MYSQL_PASSWORD: lorem
            MYSQL_DATABASE: my-app
        ports:
            - "3306:3306"
```
I also updated the Dockerfile and changed postgres_pdo into mysql_pdo, and added these:

```
FROM php:7.2-fpm-alpine3.7
RUN docker-php-ext-install mysqli pdo pdo_mysql
RUN docker-php-ext-enable mysqli
```

Unfortunately I'm still getting the following error:

`An exception occurred in driver: could not find driver
`

marcw:
You have to change the `api/config/packages/doctrine.yaml` file and change `pdo_pgsql` by `pdo_mysql` as well.

SulivanDotEu:
I'm also trying to use MySQL instead to build a proof of concept of a full reboot of our IT software and because we use MySQL, I would like to use Mysql instead of Postgres (in order to dump classes from database configuration).

I edited api/DockerFile line 17 (pdfmysql) and doctrine.yaml but it doesn't work

sushrest:
My Symfony App is still pointing to default postgres port 5432, can you point me to the right config..pls

teohhanhui:
@sushrest You also need to change the `DATABASE_URL` environment variable (in the `api/.env` file) to use the `mysql://` protocol instead of `pgsql://`

Selion05:
here's  the outpout from `git diff` with api-platform-2.3.1 and mysql as database
[git-diff.txt](https://github.com/api-platform/api-platform/files/2314225/git-diff.txt)

Selion05:
jep i guess i can do that :)

Selion05:
@dunglas  i have a guide now for the steps for 2.3.2 but i'm not very familiar with creating a pull request. 
that's my fork https://github.com/Selion05/docs/blob/2.2/core/mysql.md

Chrysweel:
[...]

PR #2130, opened 2022-02-08
Feature: optionally use mysql (but keep postgres as the default)

BurningDog:
| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | https://github.com/api-platform/api-platform/pull/1965 https://github.com/api-platform/api-platform/issues/797#issuecomment-683458553 #594 
| License       | MIT
| Doc PR        | TODO

Background: I would like to have the option to have API Platform use mysql instead of postgres. Someone made an [entirely new repo](https://github.com/pcinvent/api-platform-mysql) just to implement this feature, although it's unmaintained.

This PR adds mysql, but keeps postgres as default. It does this by:

* extracting the postgres setup to a separate `docker-compose.postgres.yml` file
* adding mysql setup to a `docker-compose.mysql.yml` file
* adding a top level `.env` file for `docker-compose` to use, which selects postgres by default (but can also be edited to use mysql instead)
* adding a variable to the Dockerfile to select with postgres or mysql
* editing the single migration file to switch between sql for postgres or mysql

To use mysql:

* edit the `.env` file, comment out the postgres lines, and uncomment the following lines:
```
# mysql
DATABASE_ENGINE=mysql
## Dev environment
COMPOSE_FILE=docker-compose.yml:docker-compose.mysql.yml:docker-compose.override.yml
```
* edit the `Dockerfile` and change `ARG DATABASE_ENGINE=postgresql` to `ARG DATABASE_ENGINE=mysql`
* run `docker-compose build`
* run `docker-compose up`

### Additional notes/concerns

* Some changes are needed to `helm/api-platform/values.yaml` but I'm not enough of a Helm/Kubernetes wizard to see what needs to happen. I compared https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml to https://github.com/bitnami/charts/blob/master/bitnami/mysql/values.yaml and tried to intuit what values would be needed, but this should be rather done by someone invested in running mysql on kubernetes.
* I haven't yet written a documentation PR, but will do once I have some feedback on this one.
* The biggest downside with this PR is that it does introduce an `.env` file for `docker-compose` to use, and adds some more `yaml` files - which makes the docker setup less intuitive. However, the big benefit is that it allows for using mysql, while maintaining the current default.
* I haven't written tests for mysql, as this would double the runtime of CI builds.
* The PR template says to update the `CHANGELOG.md` file, but I don't see one.
* This entire PRs commits need to be squashed and merge if it gets approved, as I needed to do a number of exploratory commits and test that they worked.
* re: #594 - it's possible that instead of this PR, using mysql is rather documented in the documentation repo. The benefit is less `docker-compose` complexity (and not worrying about tests not running on mysql); the downside is that it's some manual setup every time a new person wants to run mysql. I leave it to the maintainers to decide whether this PR should be solely on the docs repo.
* re: #1965 - this PR already proposed using the `.env` file, but looks abandoned. There's also a question there about moving more environment variables from the `docker-compose.yml` into the `.env` file, but I had some issues with that, so would rather first land this PR before digging in that more.

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -62,7 +62,7 @@ RUN set -eux; \
 ARG DATABASE_ENGINE=postgresql
 
 RUN if [ "$DATABASE_ENGINE" = "postgresql" ] ; then \
-    apk add --no-cache --virtual .pgsql-deps postgresql-dev; \
+	apk add --no-cache --virtual .pgsql-deps postgresql-dev; \
 	docker-php-ext-install -j$(nproc) pdo_pgsql; \
 	apk add --no-cache --virtual .pgsql-rundeps so:libpq.so.5; \
 	apk del .pgsql-deps ; \
@@ -70,7 +70,7 @@ RUN if [ "$DATABASE_ENGINE" = "postgresql" ] ; then \
 
 RUN if [ "$DATABASE_ENGINE" = "mysql" ] ; then \
 	docker-php-ext-install -j$(nproc) mysqli pdo pdo_mysql; \
-    docker-php-ext-enable mysqli ; \
+	docker-php-ext-enable mysqli ; \
 	fi
 ###< doctrine/doctrine-bundle ###
 ###< recipes ###
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -6,8 +6,6 @@
 # https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
 ARG PHP_VERSION=8.1
 ARG CADDY_VERSION=2
-ARG DATABASE_ENGINE=postgresql
-# ARG DATABASE_ENGINE=mysql
 
 # "php" stage
 FROM php:${PHP_VERSION}-fpm-alpine AS api_platform_php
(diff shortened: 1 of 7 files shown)


api/Dockerfile
```diff
@@ -59,10 +59,19 @@ RUN set -eux; \
 
 ###> recipes ###
 ###> doctrine/doctrine-bundle ###
-RUN apk add --no-cache --virtual .pgsql-deps postgresql-dev; \
+ARG DATABASE_ENGINE=postgresql
```
BurningDog, line 62: If I add this to the top of the file, then it's inaccessible at this point. I have no idea why, but it was a pretty frustrating 30 mins of trying to wrangle around it.

docker-compose.mysql.yml
```diff
@@ -0,0 +1,37 @@
...
+      volumes:
+        - db_data:/var/lib/mysql:rw
+      ports:
+        - 3306:3306
+      entrypoint:
+        - /bin/bash
+        - -c
+        - |
+          echo 'GRANT ALL on `api_test`.* to `api-platform`@`%`;' > /docker-entrypoint-initdb.d/init.sql;
+          /usr/local/bin/docker-entrypoint.sh --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
```
BurningDog, line 31: This section is so that the `api-platform` mysql user can create a test database.

dunglas:
Thanks for the contribution, however we don't plan to merge such feature. It increases the complexity and the maintenance burden. Also, Symfony Flex recipes only support Postgres by default too.
Would you mind opening a documentation PR instead?

BurningDog:
@dunglas yes, for sure - after working with this for a few days, I agree on the increased complexity - it's better to commit to either one database or the other.

I'll open a documentation PR.

PR #2153, opened 2022-03-07
fix: check for migrations in subdirectories

qooplmao:
| Q             | A
| ------------- | ---
| Branch?       | 2.6.8
| Tickets       | None
| License       | MIT

The entrypoint only checks whether there are PHP files in the ./migrations directory but they can be generated in subdirectories when using the BY_YEAR or BY_YEAR_AND_MONTH configurations.


dunglas (approved):

dunglas:
Thanks!

PR #2189, opened 2022-05-04
Fixes #1847 "Docker: Launch the php image with a non root user"

### LINKED ISSUE #1847: Docker: Launch the php image with a non root user
**API Platform version(s) affected**: 2.6.3

**Description**  
I tried to launch the php docker image with a non root user but I'm facing problems.
I added the instruction `USER www-data` at the end of the api-platform-php stage of the [dockerfile](https://github.com/api-platform/api-platform/blob/main/api/Dockerfile), (this user is created in the original php-fpm-alpine image),
but then I get permission errors with the commands executed in the [docker-entrypoint.sh](https://github.com/api-platform/api-platform/blob/main/api/docker/php/docker-entrypoint.sh#L14) file.

I tried to change the permissions of the php and composer bin, api/var files etc but it's didn't work.

**How to reproduce**  
Run docker-compose with the www-data user set in the dockerfile or in the php service of the docker-compose file.

**Additional Context**  
Configuring  and running the containers with a non root user is a requirement from or web host.
The images are meant to be running in a kubernetes environment hosted and managed by a certified health data host.

anthony-o:
| Q             | A
| ------------- | ---
| Tickets       | #1847
| License       | MIT

This PR will launch all the commands as the host user (at least, the owner id of file 'composer.json') which is convenient to be able not to generate files as "root" on the host sources : the developer using the api-platform tarball will be able to modify those files without chowning them.

One should also modify the documentation at https://api-platform.com/docs/distribution/#plugging-the-persistence-system in order to change all the `docker-compose exec` references to `docker-compose exec -T -u $(id -u)` so that all the files created will be with the host's user id and not `root`.

### DIFF
--- a/api/docker/php/docker-entrypoint.sh
+++ b/api/docker/php/docker-entrypoint.sh
@@ -1,9 +1,11 @@
 #!/bin/sh
 set -e
 
+USER_ID=$(stat -c "%u" ./composer.json)
+
 # first arg is `-f` or `--some-option`
 if [ "${1#-}" != "$1" ]; then
-	set -- php-fpm "$@"
+	set -- su-exec $USER_ID php-fpm "$@"
 fi
 
 if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
@@ -14,11 +16,12 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 	ln -sf "$PHP_INI_RECOMMENDED" "$PHP_INI_DIR/php.ini"
 
 	mkdir -p var/cache var/log
-	setfacl -R -m u:www-data:rwX -m u:"$(whoami)":rwX var
-	setfacl -dR -m u:www-data:rwX -m u:"$(whoami)":rwX var
+	setfacl -R -m u:www-data:rwX -m u:"$USER_ID":rwX var /var/run/php
+	setfacl -dR -m u:www-data:rwX -m u:"$USER_ID":rwX var /var/run/php
+	chown $USER_ID /proc/self/fd/1 /proc/self/fd/2
 
 	if [ "$APP_ENV" != 'prod' ]; then
-		composer install --prefer-dist --no-progress --no-interaction
+		su-exec $USER_ID composer install --prefer-dist --no-progress --no-interaction
 	fi
 
 	if grep -q DATABASE_URL= .env; then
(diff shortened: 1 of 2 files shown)


vincentchalamon (changes requested):
Hi @anthony-o,

Thanks for your contribution! You pointed something important related to security, it's always good to talk about it and improve it. Thanks!

For security reasons, it's recommended to run a Docker container using another user than root. Your approach only concerns some commands, and requires to pass an option on any `docker compose exec` command, which can be error prone.

I recommend to define a non-root user in the Dockerfile, by default using uid/gid 1000 (which is the most common uid/gid of the default user on many os) which can be overriden through Docker build-args, and set this user as default on the Dockerfile. Doing so will configure and run any file and command from this user, and prevent security issues.

But beware cause this can also implies some issues like permissions on volumes, etc.

api/docker/php/docker-entrypoint.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/sh
 set -e
 
+USER_ID=$(stat -c "%u" ./composer.json)
+
 # first arg is `-f` or `--some-option`
 if [ "${1#-}" != "$1" ]; then
-	set -- php-fpm "$@"
+	set -- su-exec $USER_ID php-fpm "$@"
```
dunglas, line 8: This isn't necessary. The official image is already designed to run FPM processes as non-root: https://github.com/docker-library/php/issues/70

dunglas (changes requested):
I wonder if the files shouldn't be owned by the built-in user `www-data`?

dunglas:
See also: https://github.com/Sylius/Sylius-Standard/pull/242#issuecomment-407333686

The current situation seems to be the best compromise we can make.

api/docker/php/docker-entrypoint.sh
anthony-o: Yes but I need to run the PHP script I developed as my current user ID when developing (for example to have read & write access to my folders) and by default it is running as another user id, so the files created when I execute my PHP script are not accessible by my current user ID (as a developer). This is a problem while developing... and this is why I wrote this patch.

anthony-o:
> See also: [Sylius/Sylius-Standard#242 (comment)](https://github.com/Sylius/Sylius-Standard/pull/242#issuecomment-407333686)

I agree with this comment and this is why I want to patch the entrypoint to take into account a given env variable containing the user id rather than defining it directly in the Dockerfile.

soyuka:

PR #2221, opened 2022-07-27
Add memory limit on caddy

vincentchalamon:
Backport from https://github.com/api-platform/demo/pull/259


dunglas:
According to K8S/Helm best practices, this shouldn't be enabled by default in the distribution.

PR #2231, opened 2022-08-07
Caddyfile: Use enhanced expression matcher

francislavoie:
In Caddy v2.5.2 released last month, the `expression` matcher was upgraded in https://github.com/caddyserver/caddy/pull/4715 to embed the other main built-in matchers as functions inside of the CEL expressions.

This makes it possible to more elegantly configure the `@pwa` matcher in the API Platform Caddyfile.

I didn't test this, as I don't use APIP, but I think this should be correct.

Notably, this uses the expression syntax for [`header`](https://caddyserver.com/docs/caddyfile/matchers#header) and [`path`](https://caddyserver.com/docs/caddyfile/matchers#path). In the case of the `path` matcher, it's case insensitive, so that replaces the `(?i)` in the regexp.

| Q             | A
| ------------- | ---
| Tickets       | #... 
| License       | MIT


dunglas:
For the record, this can be tested locally by just running `docker compose up`.

Then load `https://localhost` and click on the buttons allowing to access the various API Platform components.

dunglas (approved):

dunglas:
I just tested the patch, and it works! Thanks @francislavoie.

PR #2255, opened 2022-09-14
fix: react-query installation

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

We need to update this page too: https://api-platform.com/docs/main/create-client/nextjs/

### DIFF
--- a/pwa/components/common/Layout.tsx
+++ b/pwa/components/common/Layout.tsx
@@ -1,0 +1,25 @@
+import { ReactNode, useState } from "react";
+import {
+  DehydratedState,
+  Hydrate,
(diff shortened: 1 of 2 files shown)


alanpoulain (changes requested):
There is a `Layout` component doing it: https://github.com/api-platform/create-client/blob/main/templates/next/components/common/Layout.tsx

dunglas:
done (but not clean IMHO)

dunglas:
I just tried and it doesn't break the admin

PR #2259, opened 2022-09-18
Remove obsolete config files.

back-2-95:
Just remove obsolete config files.

The same configuration is already added to files in `api/config/packages` and are using keys `when@dev`, `when@test` and `when@prod`. See https://symfony.com/doc/current/configuration.html#configuration-environments


dunglas:
Thanks!

PR #2272, opened 2022-09-24
Fixed: Caddy build not working with latest dunglas/mercure|vulcain

Ninos:
I think the main branch is ok for such small "fix".

PS: Don't know if vulcain is still working (does not use it in current dev), seems there are seperate images:
https://hub.docker.com/u/dunglas

But normally vulcain should be included in mercure-image. The vulcain image got last update before two years...

Greez

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -120,20 +120,11 @@ RUN set -eux; \
 
 RUN rm -f .env.local.php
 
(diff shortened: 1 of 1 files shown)


Ninos:
Nvm, seems vulcain is not included in mercure image -> closed.

Ninos:
Ref: https://github.com/caddyserver/caddy-docker/issues/264

Issue #2294, opened 2022-10-18
Hot reload not work after client generator 


ziat1988:
**API Platform version(s) affected**: 3.0.0

**Description**  

I download the API Platform distribution then run docker on it. All the container are successfully running. Then I create some entity and generate the client Nextjs by the command 
```
docker compose exec pwa \
    pnpm create @api-platform/client --resource book -g next
```

It works. When I navigate to https://localhost I can see the page display. However, when I try to edit that book page (nextjs), nothing changes even I refresh the browser. 

I tried to navigate to http://localhost:3000, but this not works (ERR_CONNECTION_REFUSED). It seems to me that the hot reload is the problem?
Thanks

Issue #2207, opened 2022-06-20
Easy fix hot-reload for pwa


EkawaD:
**API Platform version(s) affected**: 2.6.8
**OS**: Windows 10
**Docker Desktop**: 4.9.1

**Description**  
With windows 10
Hot reloading doesn't work for pwa, it does work for the API.
I can make it work by adding WATCHPACK_POLLING=true to env vars.

**How to reproduce**  
Use this template and git clone the repo to your computer. 
Start docker desktop.
docker-compose build --pull --no-cache
docker-compose up -d 
Hot reloading doesn't work when updating pages/index.ts
Hot reloading doesn't work when updating the file pages/index.ts from the container (I opened the container in remote from vscode remote and tried changing pages/index.ts but still no hot reload)

**Possible Solution**  
I give up and used an easy fix, i just added WATCHPACK_POLLING=true to docker-compose.override.yml : 

```
...
pwa:
    build:
      context: ./pwa
      target: api_platform_pwa_dev
    volumes:
      - ./pwa:/usr/src/pwa:rw,cached
    environment:
      - WATCHPACK_POLLING=true
...
```

**Additional Context**  
There are some underlying issues with docker and windows that were already discussed [here ](https://github.com/api-platform/api-platform/issues/1816) and documented [here ](https://api-platform.com/docs/client-generator/troubleshooting/#docker-distribution-on-windows-and-hot-reloading)
It seems that it's the same issue, however it feels wrong that this easyfix isnt documented (It's windows specific and may affect performance so it shouldn't be available in the template) 
I am not sure if i should report this here but it the first place i looked after encountering this issue

soyuka:
thanks for the hint

PR #2302, opened 2022-10-25
fix(pwa): watchpack_polling to fix hot reloading on non-linux environ…

soyuka:
fixes: https://github.com/api-platform/api-platform/issues/2294 and https://github.com/api-platform/api-platform/issues/2207 and probably others


dunglas (changes requested):
This should be put in `docker-compose.override.yml` as its a dev-only setting.

dunglas:
Thanks Antoine!

PR #2404, opened 2023-03-12
remove /ping logs from php-fpm

beerfranz:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | no ticket
| License       | MIT
| Doc PR        | no doc

Since PHP-FPM 8.2, it's possible to exclude some routes to log.

https://bugs.php.net/bug.php?id=80428

I think that logging the healthcheck route `/ping` is not useful and can be remove.


dunglas (approved):

dunglas:
Thanks!

Issue #2412, opened 2023-03-22
PWA production image build is resulting in `failed to solve: invalid user index: -1`


tworzenieweb:
**API Platform version(s) affected**: 3.0.9

**Description**  
During production image build, there is an error for these two lines:

```
 > [prod 5/6] COPY --from=builder --link --chown=nextjs:nodejs /srv/app/.next/standalone ./:
 > [prod 6/6] COPY --from=builder --link --chown=nextjs:nodejs /srv/app/.next/static ./.next/static:
ERROR: failed to solve: invalid user index: -1
```

**How to reproduce**  
```
docker build -t pwa:0.1.0 -t pwa:latest pwa --target prod
```

**Possible Solution**  
Using numerical values for group and user

**Additional Context**  
The issue is related with the following docker buildx bug: https://github.com/docker/buildx/issues/1408

PR #2413, opened 2023-03-22
Fix for docker build issue for PWA

tworzenieweb:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | #2412
| License       | MIT
| Doc PR        | api-platform/docs

This PR is fixing the docker build issue with `--chown`, blocking the production image building.

### DIFF
--- a/pwa/Dockerfile
+++ b/pwa/Dockerfile
@@ -58,8 +58,8 @@ COPY --from=builder --link /srv/app/public ./public
 
 # Automatically leverage output traces to reduce image size
 # https://nextjs.org/docs/advanced-features/output-file-tracing
-COPY --from=builder --link --chown=nextjs:nodejs /srv/app/.next/standalone ./
-COPY --from=builder --link --chown=nextjs:nodejs /srv/app/.next/static ./.next/static
+COPY --from=builder --link --chown=1001:1001 /srv/app/.next/standalone ./
+COPY --from=builder --link --chown=1001:1001 /srv/app/.next/static ./.next/static
 
 USER nextjs
 


pwa/Dockerfile
```diff
@@ -58,8 +58,8 @@ COPY --from=builder --link /srv/app/public ./public
 
 # Automatically leverage output traces to reduce image size
 # https://nextjs.org/docs/advanced-features/output-file-tracing
-COPY --from=builder --link --chown=nextjs:nodejs /srv/app/.next/standalone ./
-COPY --from=builder --link --chown=nextjs:nodejs /srv/app/.next/static ./.next/static
+COPY --from=builder --link --chown=1001:1001 /srv/app/.next/standalone ./
+COPY --from=builder --link --chown=1001:1001 /srv/app/.next/static ./.next/static
```
soyuka, line 62: :-1: you can't guess that uid/gid are 1001

tworzenieweb: I'm not guessing, please consider the original buildx ticket [docker/buildx#1408](here)

dunglas:
We try to be in sync with the Docker skeleton provided by Next.js: https://github.com/vercel/next.js/blob/canary/examples/with-docker/

Could you open a PR upstream first please?

dunglas:
After a deeper look I now understand that this issue is related to the fact that we use `--link`. The fixed looks good to me. Thank you, and sorry for the long delay.

pwa/Dockerfile
dunglas: It's ok because 1001 is hardcoded in the Dockerfile: https://github.com/api-platform/api-platform/pull/2413/files#diff-c721935b4c7dd4705317d3b629d97268745c41c7bf6b47ffe185fef3fc55c2a3R54-R55

PR #2463, opened 2023-06-09
ci(pwa): use frozen PNPM lockfile

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Ensure that builds are reproducible.

### DIFF
--- a/pwa/Dockerfile
+++ b/pwa/Dockerfile
@@ -25,7 +25,7 @@ COPY --link pnpm-lock.yaml ./
 RUN pnpm fetch
 
 COPY --link . .
-RUN pnpm install -r --offline
+RUN pnpm install --offline --frozen-lockfile
 
 
 # Development image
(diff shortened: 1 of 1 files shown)


pwa/Dockerfile
```diff
@@ -34,7 +34,7 @@ FROM deps as dev
 EXPOSE 3000
 ENV PORT 3000
 
-CMD ["sh", "-c", "pnpm install -r --offline; pnpm dev"]
```
dunglas, line 37: In dev, we don't care about reproducibility nor build times.

pwa/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY --link pnpm-lock.yaml ./
 RUN pnpm fetch
 
 COPY --link . .
-RUN pnpm install -r --offline
+RUN pnpm install --offline --frozen-lockfile
```
dunglas, line 28: `-r` is useless here as we don't use workspaces, and is proven to create issues.

soyuka:
Can we use `npm run build` instead of `pnpm build` ? indeed npm also runs post/pre scripts but pnpm doesn't

soyuka (approved):

dunglas:
@soyuka the goal is to rely on PNPM only and to get rid of NPM at some point.

PR #2471, opened 2023-07-05
Fix Dockerfile last change

ili101:
First time trying to use api-platform.
I downloaded it and got the known 3.0.10 caddy build fail. But then trying main I got another error building the "pwa".
```
failed to solve: executor failed running [/bin/sh -c addgroup --system --gid 1001 nodejs        adduser --system --uid 1001 nextjs]: exit code: 1
```
Looks like last Lint conected the 2 lines without adding &&


dunglas (approved):

dunglas:
Good catch, thanks for the fix!

PR #2482, opened 2023-07-19
ci: use local buildx install

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

Docker Buildx is preinstalled in GHA runners. Let's use this instead of downloading a new one.


dunglas:
Actually, this prevents the cache from being used (this requires to share tokens etc): https://docs.docker.com/build/cache/backends/gha/

PR #2522, opened 2023-09-20
Remove duplicated COPY in Dockerfile

andersonamuller:


dunglas (approved):

dunglas:
Good catch @andersonamuller!

PR #2523, opened 2023-09-23
fix: add additional formats

divine:
| Q             | A
| ------------- | ---
| Branch?       | mai
| Tickets       | Closes https://github.com/api-platform/core/issues/5845#issuecomment-1732400657
| License       | MIT

Blank APP install visiting `https://localhost` throws an error in profiler:

```php
Uncaught PHP Exception Symfony\Component\Serializer\Exception\UnsupportedFormatException: "Serialization for the format "html" is not supported." at /app/vendor/symfony/serializer/Serializer.php line 134
```

Fix for issue with formats:

Visiting `https://localhost/docs/` doesn't throw anything.

This is the line problematic line ([source](
 https://github.com/api-platform/api-platform/blob/c7a20ed11b007a81233d6480baa5c81f9a8f7510/api/config/packages/api_platform.yaml#L11))

### DIFF
--- a/api/config/packages/api_platform.yaml
+++ b/api/config/packages/api_platform.yaml
@@ -5,6 +5,8 @@ api_platform:
     mercure: ~
     formats:
         jsonld: ['application/ld+json']
+        jsonopenapi: ['application/vnd.openapi+json']
+        html: ['text/html']
     docs_formats:
         jsonld: ['application/ld+json']
         jsonopenapi: ['application/vnd.openapi+json']


api/config/packages/api_platform.yaml
```diff
@@ -5,6 +5,8 @@ api_platform:
     mercure: ~
     formats:
         jsonld: ['application/ld+json']
+        jsonopenapi: ['application/vnd.openapi+json']
+        html: ['text/html']
```
soyuka, line 9: it should be enough to have those inside docsformats

divine: So it's a bug then? Should I close this PR?

Thanks!

vincentchalamon: `html` shouldn't be enabled by default on `formats`, otherwise it would allow to use the API with HTML contents, which is not the desired behavior.

PR #2535, opened 2023-10-03
fix: add json required format

vincentchalamon:
API Platform is unable to return a `application/problem+json` response if `application/json` format is not configured. See reproducer here: https://github.com/api-platform/demo/actions/runs/6392944298/job/17351213781


dunglas:
This is a bug in Core that must be fixed. The whole point of the error format and docs format refactoring is not to enable raw JSON by default anymore to improve security and discourage the bad practice of using JSON encoding not following a standard or an open format.

PR #2621, opened 2024-01-20
Feat : Force the local environment  to use by defaut Http

SalvadorCardona:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | n/a

When we code with defaut configuration in dev environement. Caddy force the redirection to https.

Then has several troubleshooting, the web explorer want a certificat SSL and Xdebug with https


dunglas:
Thanks for the PR, but we use HTTPS in dev on purpose.

PR #2635, opened 2024-02-15
Bug - Use Docker Composer For update-deps.sh

tstrijdhorst:
Running `update-deps.sh` failed for me since I don't have PHP or composer installed on my host system.
The other commands regarding composer are executed within the context of the php container and using that
for the recipe updates will keep everything contained to the docker environment.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/update-deps.sh
+++ b/update-deps.sh
@@ -11,8 +11,7 @@ docker compose run php /bin/sh -c 'composer update; composer outdated'
 docker compose run pwa /bin/sh -c 'pnpm install; pnpm update; pnpm outdated'
 
 # Update Symfony recipes
-cd api
-composer recipes:update
+docker compose run php /bin/sh -c 'cd api && composer recipes:update'
 
 echo 'Run `git diff` and carefully inspect the changes made by the recipes.'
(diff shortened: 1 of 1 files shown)


dunglas:
IIRC, we did that because Flex needs to access the Git repository and it isn't mounted as a volume in Docker containers (only `api/` is mounted in the `php` container).
Your patch will not work because `api/` doesn't exist in the container.

That being said, this script is mostly intended for API Platform maintainers (we may remove it from the archives at some point). Users are encouraged to use https://github.com/coopTilleuls/template-sync instead, which does much more than update the dependencies (it also updates the various config files).

tstrijdhorst:
Oh my, I could have sworn I saw it work but it definitely does not. Must be end of the day vibes, my apologies.

Thank you for taking the time to reply and pointing me in the right direction :+1:

PR #2651, opened 2024-03-04
frankenphp fmt

ili101:
Run `frankenphp fmt /srv/api/frankenphp/Caddyfile -w`
The file contained mixed tabs and spaces and trailing spaces after v3.2.14.
It's also easier to format your code when the template code is already compliant with the formatter.

### DIFF
--- a/api/frankenphp/Caddyfile
+++ b/api/frankenphp/Caddyfile
@@ -11,7 +11,7 @@
 	order php_server before file_server
 }
 
-{$CADDY_EXTRA_CONFIG}
+{$CADDY_EXTRA_CONFIG} 
 
 {$SERVER_NAME:localhost} {
 	log {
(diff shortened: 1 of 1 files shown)


api/frankenphp/Caddyfile
```diff
@@ -11,7 +11,7 @@
 	order php_server before file_server
 }
 
-{$CADDY_EXTRA_CONFIG}
+{$CADDY_EXTRA_CONFIG}
```
dunglas, line 14:
```suggestion
{$CADDY_EXTRA_CONFIG}
```

(this is a known bug in the Caddy formatter)

api/frankenphp/Caddyfile
```diff
@@ -70,13 +70,12 @@
...
-			header({'Accept': '*text/html*'})
-			&& !path(
-				'/docs*', '/graphql*', '/bundles*', '/contexts*', '/_profiler*', '/_wdt*',
-				'*.json*', '*.html', '*.csv', '*.yml', '*.yaml', '*.xml'
-			)
-		)
-		|| path('/favicon.ico', '/manifest.json', '/robots.txt', '/_next*', '/sitemap*')`
+	header({'Accept': '*text/html*'})
+	&& !path(
+	'/docs*', '/graphql*', '/bundles*', '/contexts*', '/_profiler*', '/_wdt*',
```
dunglas, line 75: Please keep the indentation as it makes the config more readable.

dunglas (changes requested):
Thanks!

api/frankenphp/Caddyfile
ili101: I didn't notice 👍

api/frankenphp/Caddyfile
ili101: I didn't notice 👍

api/frankenphp/Caddyfile
ili101: I tried to move stuff to make fmt keep the indentation but looks like it doesn't let you.
I restored it manually.

PR #2653, opened 2024-03-05
fix(doctrine): force sequence strategy

soyuka:


soyuka:
waiting for https://github.com/doctrine/migrations/issues/1406

PR #2666, opened 2024-03-18
Allow overriding TRUSTED_HOSTS and MERCURE_PUBLIC_URL variables

J3m5:
Modifying the SERVER_NAME to disable HTTPS, as outlined in the [API Platform documentation](https://api-platform.com/docs/deployment/docker-compose/#disabling-https), leads to improper configurations for TRUSTED_HOSTS and MERCURE_PUBLIC_URL when using formats like `http://localhost` or `:80`. 

This results in values like `^:80|php$` or `^http://localhost|php$` for `TRUSTED_HOSTS`, and URLs such as `https://:80/.well-known/mercure` or `https://http://localhost/.well-known/mercure` for `MERCURE_PUBLIC_URL`. 

This PR adopts "nested interpolation" from the [Docker documentation](https://docs.docker.com/compose/compose-file/12-interpolation/) to allow precise overrides, introducing `TRUSTED_PROXIES` and `CADDY_MERCURE_PUBLIC_URL` as new, adjustable variables, with room for naming discussions.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -37,7 +37,7 @@ services:
     environment:
       NEXT_PUBLIC_ENTRYPOINT: http://php
 
-  ###> doctrine/doctrine-bundle ###
+###> doctrine/doctrine-bundle ###
   database:
     image: postgres:${POSTGRES_VERSION:-16}-alpine
     environment:
@@ -58,7 +58,7 @@ services:
 volumes:
   caddy_data:
   caddy_config:
-  ###> doctrine/doctrine-bundle ###
+###> doctrine/doctrine-bundle ###
   db_data:
 ###< doctrine/doctrine-bundle ###
 ###> symfony/mercure-bundle ###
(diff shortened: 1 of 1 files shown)


compose.yaml
```diff
@@ -37,7 +37,7 @@ services:
     environment:
       NEXT_PUBLIC_ENTRYPOINT: http://php
 
-###> doctrine/doctrine-bundle ###
+  ###> doctrine/doctrine-bundle ###
```
vincentchalamon, line 40:
```suggestion
###> doctrine/doctrine-bundle ###
```

compose.yaml
```diff
@@ -58,7 +58,7 @@ services:
 volumes:
   caddy_data:
   caddy_config:
-###> doctrine/doctrine-bundle ###
+  ###> doctrine/doctrine-bundle ###
```
vincentchalamon, line 61:
```suggestion
###> doctrine/doctrine-bundle ###
```

vincentchalamon (approved):
LGTM

dunglas (approved):
LGTM. Could you update the docs to mention these new variables (the current example is still broken, even with your changes, because the new variables must be explicitly set) and also update Symfony Docker accordingly? https://github.com/dunglas/symfony-docker/blob/main/compose.yaml

J3m5:
> LGTM. Could you update the docs to mention these new variables (the current example is still broken, even with your changes, because the new variables must be explicitly set)

Of course.

In which branch this change will be merged?
I suppose I'll need to handle the documentation with the new variables and without in previous branches, right @dunglas ?
How do I should proceed?

>  and also update Symfony Docker accordingly? https://github.com/dunglas/symfony-docker/blob/main/compose.yaml

Done: https://github.com/dunglas/symfony-docker/pull/591

dunglas:
Target the 3.3 branch for the docs. We don't have branches for the distribution.

PR #2710, opened 2024-05-26
fix: serve static files from pwa if not found in api

ibousfiha:
Hi, this PR is to serve static files like svgs from pwa if not found in api.
Please let me know if it fits or updates are required.

| Q             | A
| ------------- | ---
| Branch?       | main 
| Tickets       | Closes #..., closes #... 
| License       | MIT


dunglas:
Good idea, but this will prevent the API from serving files with this extension if any.

We should use `try_files` to first check if the file exists API-side, and redirect to the PWA if it's not the case. This will have the added benefit of working with any file.

soyuka:

PR #2747, opened 2024-07-27
Facilitate admin customization

### LINKED ISSUE #2745: Make initial admin customization easier
**Description**  

When developers want to customize the admin, it's not super straightforward because the admin is in a dynamic import:

https://github.com/api-platform/api-platform/blob/c5240c2f4ecf43d5b197de12c80a33b20086134b/pwa/pages/admin/index.tsx#L7-L13

In order to customize the admin, the developer has to extract this into another component, then start cusomizing the component. 

A better solution would be to extract the admin app by default:

```jsx
// in pwa/pages/admin/index.tsx
import Head from "next/head";
import { useEffect, useState } from "react";
import styles from "./admin.module.css";

const Admin = () => {
  // Load the admin client-side
  const [DynamicAdmin, setDynamicAdmin] = useState(<p>Loading...</p>);
  useEffect(() => {
    (async () => {
      const App = (await import("../../components/admin/App")).App;
      setDynamicAdmin(<App />);
    })();
  }, []);

  return (
    <div className={styles.admin}>
      <Head>
        <title>API Platform Admin</title>
      </Head>

      {DynamicAdmin}
    </div>
  );
};
export default Admin;
```

and provide a ready-to-customize `App` component without lazy loading:

```jsx
// in pwa/components/admin/App.tsx
import { HydraAdmin } from "@api-platform/admin";

export const App = () => <HydraAdmin entrypoint={window.origin}></HydraAdmin>;
```

fzaninotto:
Closes #2745

### DIFF
--- a/pwa/pages/admin/index.tsx
+++ b/pwa/pages/admin/index.tsx
@@ -1,25 +1,12 @@
-import Head from "next/head";
-import { useEffect, useState } from "react";
+import type { NextPage } from "next";
+import dynamic from "next/dynamic";
 
-const Admin = () => {
-  // Load the admin client-side
-  const [DynamicAdmin, setDynamicAdmin] = useState(<p>Loading...</p>);
-  useEffect(() => {
-    (async () => {
-      const HydraAdmin = (await import("@api-platform/admin")).HydraAdmin;
+// load the admin client-side
+const App = dynamic(() => import("../../components/admin/App"), {
+  ssr: false,
+  loading: () => <p>Loading...</p>,
+});
 
-      setDynamicAdmin(<HydraAdmin entrypoint={window.origin}></HydraAdmin>);
-    })();
-  }, []);
+const Admin: NextPage = () => <App />;
 
-  return (
-    <>
-      <Head>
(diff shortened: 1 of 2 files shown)


pwa/pages/admin/index.tsx
```diff
@@ -1,25 +1,12 @@
...
+import dynamic from "next/dynamic";
 
-const Admin = () => {
-  // Load the admin client-side
-  const [DynamicAdmin, setDynamicAdmin] = useState(<p>Loading...</p>);
-  useEffect(() => {
-    (async () => {
-      const HydraAdmin = (await import("@api-platform/admin")).HydraAdmin;
+// load the admin client-side
+const App = dynamic(() => import("../../components/admin/App"), {
```
fzaninotto, line 5: This is the idiomatic way to lazy load a client-side component in next.js (and also the advised way to integrate react-admin and next.js, see https://marmelab.com/react-admin/NextJs.html)

dunglas:
Thanks @fzaninotto!

PR #2761, opened 2024-09-09
feat: Refine CI docker cache

nervo:
Refine CI docker cache handling, and obtain 100% cache hit during containers building.

While playing with [docker bake action](https://github.com/docker/bake-action) and github action based docker build layer cache, i've struggled a bit to discover that, in case of a multi target baking,  *every* target *must* have its own scope (see: https://github.com/docker/bake-action/issues/91#issuecomment-1265817118).

As this repo uses 2 targets, i've just used them as a prefix for each target cache scope:
```shell
$ docker buildx bake -f compose.yaml -f compose.override.yaml --print
...
{
  "group": {
    "default": {
      "targets": [
        "php",
        "pwa"
      ]
    }
  },
...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,9 +32,12 @@ jobs:
             compose.yaml
             compose.override.yaml
           set: |
-            *.cache-from=type=gha,scope=${{github.ref}}
-            *.cache-from=type=gha,scope=refs/heads/main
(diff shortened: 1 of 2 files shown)


dunglas (approved):

dunglas:
Awesome! Thanks for the research and for the fix!

PR #2844, opened 2025-02-09
Add port binding for PWA container

aruku:
My browser can't reach the client app without this.


rvanlaak:
If I understand correctly; Caddy should proxy all requests (see Caddyfile), so a direct connection to the PWA client should not be needed.

aruku:
You are right, this was probably caused by some Docker weirdness from upping and downing multiple containers.

PR #2868, opened 2025-03-28
Fix Docker build FromAsCasing mismatch warning

rvanlaak:
FromAsCasing mismatch on AS and FROM keywords.

Docker build output:

image

| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | -
| License       | MIT
| Doc PR        | -


rvanlaak:
Happy to resolve them with this PR, but this template repo is not that active so will need some maintainer support in getting efforts merged as well.

PR #2922, opened 2025-08-08
Fix Dockerfile ENV legacy usage

guillemfondin:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | 
| License       | MIT
| Doc PR        | N/A

Fix this warning message when docker build pwa image :
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format


dunglas (approved):

dunglas:
Thanks!

PR #2926, opened 2025-09-01
feat: add health checks to prevent 502 in CI

dunglas:
| Q             | A
| ------------- | ---
| Branch?       | main
| Tickets       | n/a
| License       | MIT
| Doc PR        | n/a

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -31,13 +31,23 @@ services:
       - target: 443
         published: ${HTTP3_PORT:-443}
         protocol: udp
-
+    healthcheck:
+      test: curl --insecure --fail https://localhost/docs || exit 1
+      timeout: 5s
+      retries: 5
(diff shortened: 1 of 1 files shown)


rvanlaak:
The Dockerfile [already has a healthcheck defined](https://github.com/api-platform/api-platform/blob/d0dbde7513fa1fd5d71adf71c7429061c05c1ab0/api/Dockerfile#L56), pointing to the Caddy metrics: https://caddyserver.com/docs/metrics#caddys-metrics

Is this approach to query the docs (which are generated through php) preferred over the use of the Caddy metrics? What about moving this to the `frankenphp_dev` Docker layer then though?

dunglas:
Good catch. It was to fix a CI failure but we should definitely uniformize.

Issue #3008, opened 2026-08-24
Outdated symfony-docker


ili101:
I was confused about why I saw files that are not in this project or in https://github.com/dunglas/symfony-docker/ in the created project, but then I saw that this is locked to a specific commit on symfony-docker and that the files are from there:
https://github.com/api-platform/api-platform/blob/ecce0af6a0ae23d6501ba301021cb6d3e7bf3761/src/Scaffold/SymfonyScaffold.php#L18
Can we update this to the latest commit, or maybe point to "main" instead? What do you think?

Issue #2994, opened 2026-06-19
Migration from the old template to this


ili101:
I am using the old version of the template (with v3.3.6), and I wanted to try 2 things: create a new project, and then potentially "update" my existing project.
Following the direction verbatim didn't go well:
<img width="979" height="1143" alt="Image" src="https://github.com/user-attachments/assets/46888373-47c6-4179-9018-2475ded8369b" />

So I opened VSCode, created a new PHP 8.5 devcontainer to bootstrap the installer and tried again in it, but I got Docker and Node  missing errors, so I added them to the config:
```json
{
	"name": "PHP",
	"image": "mcr.microsoft.com/devcontainers/php:3-8.5-trixie",
	"features": {
		"ghcr.io/devcontainers/features/node:2": {},
		"ghcr.io/postfinance/devcontainer-features/docker-out:1": {}
	}
}
```
Now downloading and running it again worked (gave some build error but completed):
```
vscode ➜ /workspaces/temp1 $ ./api-platform my-app --framework=symfony --with-docker --with-pwa --with-admin
...
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: sharp@0.34.5, unrs-resolver@1.12.2

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.

Aborting installation.
  pnpm install has failed.

Approving required pnpm build scripts
$ 'pnpm' 'install'
✓ Lockfile passes supply-chain policies (verified 655ms ago)
Lockfile is up to date, resolution step is skipped
Already up to date
node_modules/.pnpm/unrs-resolver@1.12.2/node_modules/unrs-resolver: Running postinstall script, done in 40ms
node_modules/.pnpm/sharp@0.34.5/node_modules/sharp: Running install script, done in 67ms
Done in 1.2s using pnpm v11.8.0
Installing API Platform frontend libraries
$ 'pnpm' 'add' '@api-platform/api-doc-parser' 'github:api-platform/zod' '@api-platform/ld' '@api-platform/mercure'
✓ Lockfile passes supply-chain policies (verified 1s ago)
[WARN] 1 deprecated subdependencies found: node-domexception@1.0.0
Packages: +28
++++++++++++++++++++++++++++
Progress: resolved 458, reused 359, downloaded 28, added 28, done

dependencies:
+ @api-platform/api-doc-parser 0.16.10
+ @api-platform/ld 1.0.0
+ @api-platform/mercure 1.0.0
[...]

PR #3009, opened 2026-08-25
feat: add database option to installer. updated: SYMFONY_DOCKER_REF

ili101:
Added `--with-database` issue https://github.com/api-platform/api-platform/issues/2994#issuecomment-5401632688. Merge if you think it's beneficial.

Updated SYMFONY_DOCKER_REF to the latest commit (rn https://github.com/dunglas/symfony-docker/commit/bd4320fa303d506c8a1d0593f0e9a7ed45dc13fb) issue https://github.com/api-platform/api-platform/issues/3008. Or we can use "main" if you think it's preferable.


soyuka:
I don't think this is useful wdyt @dunglas ?

dunglas:
I don't think either. You just need to type `composer require orm` thanks to flex!

